PR #14, opened 2026-08-19
Add nightly runtime pre-release workflow

mordamax:
## Changes

- Extract the shared srtool runtime build into a reusable workflow (`build-runtimes.yml`) used by both the manual `v*` release and the new nightly.
- Add `nightly-release.yml` publishing the two paseo runtimes as GitHub pre-releases on a nightly schedule (and on-demand), so downstream CI can test the newest runtime features without waiting for a versioned release.
- Each nightly pre-release carries a short diff of the PRs included since the previous nightly.


.github/workflows/build-runtimes.yml
```diff
@@ -0,0 +1,43 @@
+name: Build runtimes
+
+# Shared srtool build for the two paseo runtimes. Called by the manual Release
+# and the nightly pre-release workflows so both ship identical artifacts.
+on:
+  workflow_call:
+
+jobs:
+  build-runtimes:
+    runs-on: parity-int-ubuntu
```
andrzejSulkowski: not sure, but are the `int` runners 'internal' parity runners? they seemed not to work when setting ci/cd in this public repo.

.github/workflows/build-runtimes.yml
```diff
@@ -0,0 +1,43 @@
...
+          - name: next-asset-hub-paseo
+            package: next-asset-hub-paseo-runtime
+            path: runtimes/next-asset-hub-paseo
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
+
+      - name: Build ${{ matrix.name }} runtime
+        id: srtool_build
+        uses: chevdor/srtool-actions@v0.9.2
```
andrzejSulkowski, line 26: we have a `zombienet-snapshot.yml` workflow running which as well builds the runtimes and runs the initialization scripts on a zombienet, snapshots the dbs and publishes them.

Maybe building the runtime could be done only once and shared between this and the zombienet snapshot.

.github/workflows/build-runtimes.yml
mordamax: good catch)

.github/workflows/build-runtimes.yml
mordamax: these are different builds though - release needs srtool so the wasm is verifiable, zombienet builds with plain cargo on purpose so the snapshot matches what people build locally. if we reused the srtool wasm there, fetch-snapshot would warn about drift for everyone

andrzejSulkowski (approved):

.github/workflows/build-runtimes.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          chain: ${{ matrix.name }}
+          package: ${{ matrix.package }}
+          runtime_dir: ${{ matrix.path }}
+          profile: production
+
+      - name: Store srtool digest
```
csmnprschv, line 33: Make sure to apply #12's fix here.

csmnprschv (approved):

.github/workflows/build-runtimes.yml
mordamax: 19e4607 done thanks

PR #19, opened 2026-08-20
fix(people-lite): reject alias accounts in attest

### LINKED ISSUE #5: People Lite: `attest` call AccountToAlias gate
The `attest` call does not gate on accounts which are already mapped to an alias.

include:

```rs
ensure!(!AccountToAlias::<T>::contains_key(&candidate), Error::<T>::AccountInUse);
```

agustinustheo:
Closes #5

`attest` only rejected accounts registered as lite people, so an account already bound to a lite alias could also become a canonical lite person.

The call now rejects alias-bound candidates with `AccountInUse` before registration. A regression test establishes a real alias binding and verifies that attestation leaves the binding and allowance unchanged.

## Changes

- Add the `AccountToAlias` gate to `PeopleLite::attest`.
- Add a regression test for an alias-bound candidate.

### DIFF
--- a/pallets/people-lite/src/lib.rs
+++ b/pallets/people-lite/src/lib.rs
@@ -559,7 +559,7 @@ pub mod pallet {
 				T::AccountContexts::contains(&rev_ca.ca.context),
 				Error::<T>::InvalidAliasContext
 			);
-			ensure!(!LitePeople::<T>::contains_key(&account), Error::<T>::AccountInUse);
+			ensure!(!LitePeople::<T>::contains_key(&account), Error::<T>::AlreadyRegistered);
 
 			let old_account = AliasToAccount::<T>::get(&rev_ca.ca);
 			let old_rev_ca = old_account.as_ref().and_then(AccountToAlias::<T>::get);
--- a/pallets/people-lite/src/lib.rs
+++ b/pallets/people-lite/src/lib.rs
@@ -461,6 +461,7 @@ pub mod pallet {
 		) -> DispatchResult {
 			let candidate = ensure_signed(origin)?;
 			ensure!(!LitePeople::<T>::contains_key(&candidate), Error::<T>::AlreadyRegistered);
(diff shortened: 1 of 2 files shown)


pallets/people-lite/src/lib.rs
```diff
@@ -377,6 +377,7 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			let verifier = ensure_signed(origin)?;
 			ensure!(!LitePeople::<T>::contains_key(&candidate), Error::<T>::AlreadyRegistered);
+			ensure!(!AccountToAlias::<T>::contains_key(&candidate), Error::<T>::AccountInUse);
```
csmnprschv, line 380: Add it to `register_with_fee` as well.

csmnprschv (approved):

agustinustheo:
I force pushed because I rebased with `main`.

pallets/people-lite/src/lib.rs
agustinustheo: Added: https://github.com/paritytech/individuality-community/pull/19/commits/273ea5f4f73dd566d88a66fc56b97734b8caff6b

andrzejSulkowski (approved):

pallets/people-lite/src/lib.rs
```diff
@@ -377,6 +377,7 @@ pub mod pallet {
 		) -> DispatchResultWithPostInfo {
 			let verifier = ensure_signed(origin)?;
 			ensure!(!LitePeople::<T>::contains_key(&candidate), Error::<T>::AlreadyRegistered);
```
tindzk, line 379: In `set_alias_account`, a different error code is used for the same condition:

```rust
ensure!(!LitePeople::<T>::contains_key(&account), Error::<T>::AccountInUse);
```

agustinustheo: Fixed: https://github.com/paritytech/individuality-community/pull/19/commits/37c22de1d2840e760997f1307b790ff1c6ac7559

PR #10, opened 2026-08-19
update to polkadot-sdk 2606

kianenigma:
It will eventually be needed 🤫

Main review points: 

* sp_core hashing re-exports removed (#12158) — migrated twox_64/twox_256/blake2_256 usages in 9 pallets + the people runtime to sp_crypto_hashing, adding the workspace dep and per-crate std feature wiring.
* frame_support::error module removed (#12149) — frame_support::error::BadOrigin → sp_runtime::traits::BadOrigin in pallets/members.
* Candidate descriptor v3 (#10742) — in both runtimes and the relay-randomness mock:
  * type SchedulingSignatureVerifier = (); in cumulus_pallet_parachain_system::Config (v3 scheduling disabled, matching asset-hub/people-westend)
  * fn max_claim_queue_offset() in RelayParentOffsetApi (delegates to the pallet)
  * new SchedulingV3EnabledApi runtime API

PR #21, opened 2026-08-20
bump pallet-revive to 0.18.1

re-gius:
DotNS deploys onto next-asset-hub-paseo, and its PoP controller has to recognise a substrate Root dispatch as an authority. The only root check in pallet-revive 0.18.0 is callerIsRoot, which answers only when the asking contract is the direct callee of the dispatch. The controller is a UUPS proxy, so its implementation runs inside the proxy's delegatecall and cannot ask the precompile from its own frame.

pallet-revive 0.18.1 adds originIsRoot, which returns true whenever the top-level dispatch was root regardless of how many contract or delegate-call frames separate the precompile from that dispatch. That lets the controller check root authority from inside its own proxy frame and it is the same mechanism any future governance-gated entrypoint on the registrar controller would use. The method was published from unstable2604 after the backport of https://github.com/paritytech/polkadot-sdk/pull/12281 landed as https://github.com/paritytech/polkadot-sdk/pull/12923. This is a lockfile-only update and it makes the new available to next-asset-hub-paseo, which has to ship for the method to exist on chain.

## Changes

- Cargo.lock: the three version bumps for pallet-revive, pallet-revive-uapi, pallet-revive-fixtures.
- Five transitive windows-sys edges re-point from 0.61.2 to 0.59.0 as a resolver side effect


csmnprschv (approved):

csmnprschv:
@kianenigma, we'll need to revisit `pallet-revive` when we take care of #10. It uses `0.19.0`, which doesn't currently contain `originIsRoot`.

re-gius:
> @kianenigma, we'll need to revisit `pallet-revive` when we take care of #10. It uses `0.19.0`, which doesn't currently contain `originIsRoot`.

If necessary, I can backport https://github.com/paritytech/polkadot-sdk/pull/12281 to some other (stable?) branch to have it in the ^0.19.0 releases

csmnprschv:
> > @kianenigma, we'll need to revisit `pallet-revive` when we take care of #10. It uses `0.19.0`, which doesn't currently contain `originIsRoot`.
> 
> If necessary, I can backport [paritytech/polkadot-sdk#12281](https://github.com/paritytech/polkadot-sdk/pull/12281) to some other (stable?) branch to have it in the ^0.19.0 releases

Yeah, would be nice if the upcoming `polkadot-stable2606-2` had it.

re-gius:
> Yeah, would be nice if the upcoming `polkadot-stable2606-2` had it.

backport in progress paritytech/polkadot-sdk#12959

gui1117 (approved):

PR #27, opened 2026-08-21
nft-credits: Build one credit tree per proof query

tindzk:
`nft_claim_credit_proofs` rebuilt the whole Merkle tree for each of a claimant's credits in an award block, so a query cost `n + m*(2n-1)` hashes and cloned the leaf set `m` times. The runtime API is reachable over `state_call`, which no node meters.

The tree is now walked once per query, collecting the siblings of every requested index as the layers are hashed. At `MaxCreditsPerBlock = 1200` and 15 credits per claimant, that is ~2 400 blake2_256 calls instead of ~37 000, and ~5 KB copied instead of ~576 KB. The remaining cost is one tree build.

Fixes https://intelligence.security.parity.io/finding/individuality-1249-1786709755/6009


Zebedeusz (approved):

pallets/nft-credits/src/lib.rs
```diff
@@ -1151,14 +1155,14 @@ impl<T: Config> Pallet<T> {
...
 			.enumerate()
 			.filter(|(_, award)| &award.claimant == claimant)
-			.map(|(leaf_index, award)| {
-				Self::credit_proof(&recorded, &leaves, award.credit, leaf_index as u32)
-			})
-			.collect::<Result<Vec<_>, _>>()
+			.map(|(leaf_index, award)| (leaf_index as u32, award.credit))
+			.collect::<Vec<_>>();
+
+		Self::credit_proofs(&recorded, &leaves, &claimed)
```
agustinustheo: Looking at this, I think it's better to return early when there is no credits in this block.

Because with the current code it will still rebuild the complete tree even when `claimed` is empty.

```suggestion
		let claimed = awards
			.iter()
			.enumerate()
			.filter(|(_, award)| &award.claimant == claimant)
			.map(|(leaf_index, award)| (leaf_index as u32, award.credit))
			.collect::<Vec<_>>();
		if claimed.is_empty() {
			return Ok(Vec::new());
		}

		let leaves = Self::nft_claim_credit_leaves(&awards);
		Self::credit_proofs(&recorded, &leaves, &claimed)
```

pallets/nft-credits/src/lib.rs
```diff
@@ -1137,6 +1137,10 @@ impl<T: Config> Pallet<T> {
 	/// Exposed through the runtime API rather than as a call: nothing is written, and a whole
 	/// block's awards in an extrinsic's proof would be paid for by every other extrinsic in
 	/// the block.
+	///
+	/// One call hashes the block's leaf set once, however many proofs it returns, so the cost is
+	/// bounded by [`Config::MaxCreditsPerBlock`] and does not grow with the credits the claimant
+	/// holds.
```
agustinustheo: If my previous comment was implemented (https://github.com/paritytech/individuality-community/pull/27/changes#r3850388174), this might need some changes as well.

```suggestion
	/// A call with at least one proof derives the leaves and builds the block's tree once.
	/// [`Config::MaxCreditsPerBlock`] bounds the tree hash count, regardless of the proof count.
	/// Each proof adds collection work and response data.
```

agustinustheo (approved):
Please review the comments before merging.

PR #28, opened 2026-08-21
game: Clean up Merkle trees

tindzk:
Until now, credit trees were retained indefinitely on both chains. This PR implements removal:

- A tree is removed once its credits are minted or its claim deadline has passed
- Each chain (People, Asset Hub) tells the other what it removed
- Each chain has a TTL sweep of its own that covers for a message that never arrived

## Lifecycle of one tree

```mermaid
stateDiagram-v2
    state "Fully claimed" as Claimed
    [*] --> Stored: receive_credit_trees
    [*] --> Rejected: already past its deadline on arrival
    Stored --> Claimed: every leaf claimed
    Stored --> Expired: claim deadline passed, swept by bucket
    Claimed --> Queued: deletion queued
    Expired --> Queued: deletion queued
    Queued --> [*]: sent to the People chain
    Rejected --> [*]
```

## Delivery and deletion across the two chains

```mermaid
sequenceDiagram
    participant PeopleOcw as People offchain worker
    participant Credits as nft-credits, People
    participant Claims as nft-claims, Asset Hub
    participant AhOcw as Asset Hub offchain worker
    participant User as Claimant

    Credits->>Claims: receive_credit_trees
    User->>Claims: claim, mints the NFT
    Note over Claims: last leaf of a tree claimed,<br/>tree removed and deletion queued
    AhOcw->>Claims: sweep_expired_trees, one bucket
    Note over Claims: trees past the claim deadline removed,<br/>deletions queued
    AhOcw->>Claims: send_tree_deletions
    Claims->>Credits: receive_tree_deletions, XCM
    Note over Credits: named roots removed
    PeopleOcw->>Credits: sweep_expired_roots, one bucket
    Note over Credits: roots past their own deadline removed,<br/>covers for a lost message
```

Both sweeps walk one bucket of wall-clock time per call. A tree's bucket is the UTC day its credits
were awarded in. A sweep therefore selects the trees that are due instead of reading every tree.

On Asset Hub, a tree is dropped 90 days after the credits were awarded, which is the claim deadline.
The People chain keeps the root it was built from for a further 30 days. That deadline is derived
from the Asset Hub one, so the two cannot drift apart. A credit inside its deadline always has a root
and its awards to prove against.

## Changes

- **nft-claims**: removes a tree once it is fully claimed and once its claim deadline has passed. A
  tree that arrives past its deadline is not stored. Removals queue up and go out to the People chain
  as one message, both driven by the offchain worker.
- **nft-credits**: removes the roots Asset Hub reports it is finished with. It sweeps the ones no
  message ever named.
- **indiv-support**: holds the expiry-bucket scheme and the sweep helpers both chains drive. The
  offchain submission is shared too, so a stalled submitter behaves the same on either chain.
- **Runtimes**: the People chain accepts Asset Hub's unpaid deletion message. Each runtime's
  integrity test asserts that its sweeps fit a block and that a full deletion message fits the
  channel.


Zebedeusz (approved):

pallets/nft-claims/src/lib.rs
```diff
@@ -816,6 +1209,314 @@ pub mod pallet {
...
+			)
+				.encode();
+
+			send_xcm::<T::XcmRouter>(T::GameChainLocation::get(), Self::tree_deletion_xcm(call))
+				.map(|_| ())
+		}
+
+		fn tree_deletion_xcm(encoded_call: Vec<u8>) -> Xcm<()> {
+			Xcm(vec![
+				UnpaidExecution { weight_limit: WeightLimit::Unlimited, check_origin: None },
```
andrzejSulkowski, line 1439: why `WeightLimit:Unlimited` and not a benchmarked weight?

shawntabrizi (approved):
This code may need a storage migration depending on when it lands.

Otherwise stale storage will mess up the state transition function.

pallets/nft-claims/src/lib.rs
tindzk: `Limited(m)` must cover the weight the destination (game chain) computes. We cannot benchmark it here. A hardcoded value would silently drop messages if the game chain weights change.

support/src/offchain.rs
```diff
@@ -0,0 +1,89 @@
...
+/// Longevity, in blocks, of an offchain worker's submission.
+/// A stranded retry leaves the pool instead of waiting to be mined against state it no longer
+/// matches.
+pub const TX_LONGEVITY: u64 = 64;
+
+/// Period, in blocks, at which a failed submission logs a warning instead of a `debug` message.
+///
+/// A submission fails inside every retry window, so only a streak of failures shows a stall.
+/// Nothing else reports a stalled submitter.
+pub const STALL_WARN_PERIOD: u32 = 32;
```
gui1117: is it true that it is only a streak of failures? what if the block number of the first submission is a multiple of `STALL_WARN_PERIOD`?

I think we can also just always warn, and if it happens so regularly we can reconsider the code.

pallets/nft-claims/src/lib.rs
```diff
@@ -271,31 +412,61 @@ pub mod pallet {
...
+
+	/// The oldest bucket of [`TreeExpiries`] that has not been swept.
+	///
+	/// `None` until the first tree is stored. A sweep would otherwise start at the UNIX epoch and
+	/// walk one bucket per call to reach the present. A tree that arrives in an earlier bucket
+	/// than any held so far lowers it, which an out-of-order delivery does. Such a tree cannot
+	/// belong to a bucket already swept, because that bucket was due in full and this chain
+	/// rejects a tree past its deadline.
+	#[pallet::storage]
+	pub type NextExpiryBucket<T: Config> = StorageValue<_, ExpiryBucket, OptionQuery>;
```
gui1117: if you wanted you could simplify by using a Identity hasher for TreeExpiries first key, and store ExpiryBucket number in BigEndian.

Then to get the next expiring bucket you would just do `TreeExpiries::next()`.

You can even remove the concept of bucket entirely and store `AwardBlock` in BigEndian. And then sweep by iterating until recent one or a maximum.
And same in pallet nft-claims.

But the current implementation works.

pallets/nft-claims/src/lib.rs
```diff
@@ -529,14 +784,27 @@ pub mod pallet {
...
+						.saturating_add(T::Nfts::mint_hook_weight(CLAIM_METADATA_PAIRS)),
+				)
+				.into());
+			}
+
+			// No proof can be built against a fully claimed tree again, so remove it and tell the
+			// game chain to drop its root. The spent leaves stay, because a replay there can
+			// deliver the tree again before the deletion arrives, and only those leaves keep its
+			// credits spent.
+			Self::remove_tree(block);
```
gui1117, line 787: it removes one entry on this chain, but 2 are kept, and 88 Bytes removed, 252 Bytes kept.

So not sure if it is worth it that much. That said it removes every thing on PC, but PC state size is less important.

EDIT: but also it is ok.

support/src/offchain.rs
tindzk: Good point. It holds as long as `STALL_WARN_PERIOD` is a multiple of `RETRY_WINDOW`.

I will improve the design.

pallets/nft-claims/src/lib.rs
tindzk: The current approach should be fine since the entries will be eventually removed by the sweep.

pallets/nft-claims/src/lib.rs
tindzk: Good idea. This simplifies the implementation. I have made these changes.

PR #33, opened 2026-08-21
Fix game `offboard` bypassing personhood suspension

georgepisaltu:
Fixes an issue where users who would get recognized as people in `pallet-score` could then call `offboard` and, because they were not present in `pallet-game` (our only `pallet-score` consumer), there would be no runtime logic to ever suspend them even as they would not be playing any games.

This PR fixes this by suspending recognized (but not externally recognzied) people when they offboard in a special mutation session. It's acceptable to run this mutation session and therefore to rebuild the ring because it is a really expensive attack (it's difficult to get your personhood back if you are suspended) so it doesn't scale.


pallets/game/src/lib.rs
gui1117, line 1341: If it happens that a user wants to restart with the same key after being kicked out, then we maybe we can implement a way to resume from being kicked out.

But probably better to never kick out. Kicking out made sense when user could sign up with a very low deposit. If we only accept lite-people then I think kicking out is not useful anymore.

gui1117 (approved):
Thank you.

In the future, if we remove the sign up with the cheap deposit then we can remove `offboard` and `kickout`. Ideally we should only accept lite-people and we never need to kick them out.

pallets/game/src/lib.rs
georgepisaltu: Maybe we merge this one and then we remove the kick after we confirm we don't need the small deposit sign up anymore.

pallets/game/src/lib.rs
```diff
@@ -1252,6 +1252,11 @@ pub mod pallet {
 		///
 		/// There must be no game or the existing game must be in registration phase and the player
 		/// must have not signed up for the game.
+		///
+		/// A player whose pallet-score recognition is `Recognized` has their personhood suspended,
+		/// the same as an absence-triggered suspension. Offboarding is permanent. Playing as a
+		/// person again after a suspension and then re-recognition requires onboarding under a new
```
andrzejSulkowski, line 1258: Did we communicate to App engineers that a `kickout` will permanently burn the key?

andrzejSulkowski (approved):

PR #38, opened 2026-08-24
pallet-members: merge-rings is spammable. 

gui1117:
# Issue

merge-rings can merge with empty rings and beyond the `current_ring_index` of the collection. One spam is to just merge half-removed ring into the index 100000001 then merge 100000001 into 100000002 etc... and merges are refunded on success so this is free. (And this even triggers the subscriber notifier process.)

# Fix

New additional condition to merge rings: both are non-empty.

Also there is another small unrelated issue with merged-away rings leaving empty pages but it is fixed in https://github.com/paritytech/individuality-community/pull/39

### DIFF
--- a/pallets/members/src/tests.rs
+++ b/pallets/members/src/tests.rs
@@ -1053,6 +1053,287 @@ mod merge_rings_tests {
 	}
 }
 
+/// `merge_rings` is a free call available to any signed account, so it must never accept ring
+/// indices that do not refer to two existing, populated rings. Otherwise it can be spammed to
+/// write junk into arbitrary ring indices, or to shuffle a real ring around the index space at no
+/// cost.
+mod merge_rings_spam_tests {
+	use super::*;
+
+	/// Keep building `ring_index` until every one of its keys is included in its root.
+	fn finish_ring_build(identifier: Identifier, ring_index: RingIndex) {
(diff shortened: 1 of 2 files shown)


pallets/members/src/tests.rs
```diff
@@ -1053,6 +1053,288 @@ mod merge_rings_tests {
...
+			for (base, target) in [(2u32, 0u32), (2, 1), (2, 4)] {
+				assert_noop!(
+					MembersPallet::merge_rings(RuntimeOrigin::signed(1), identifier, base, target),
+					Error::<Test>::RingAboveMergeThreshold
+				);
+			}
+			// Ring 3 is the top ring and the collection never created a ring 4; neither of them is
+			// a merge partner.
+			for (base, target) in [(3u32, 2u32), (2, 3), (0, 3), (3, 0), (4, 2)] {
+				dbg!(base, target);
```
andrzejSulkowski: probably left over.

andrzejSulkowski (approved):

georgepisaltu (approved):

pallets/members/src/tests.rs
gui1117:
```suggestion
```

gui1117: yes fixed

PR #44, opened 2026-08-24
Fix leak of sufficient reference on personhood suspension

georgepisaltu:
People have an account tied to their personhood which gets a sufficient reference when registering as a person. When a person is suspended, we remove the account link from storage but we don't take the sufficient reference back, but this PR fixes this.

With the latest developments in this project, maybe it's worth removing the associated account altogether since I can't think of a use case for it now.


pallets/people-multi/src/lib.rs
```diff
@@ -972,9 +972,9 @@ pub mod pallet {
 			let mut keys = vec![];
 			for personal_id in suspensions {
 				let mut record = People::<T>::get(personal_id).ok_or(Error::<T>::NotPerson)?;
-				if let Some(account) = record.account {
-					AccountToPersonalId::<T>::remove(account);
-					record.account = None;
+				if let Some(account) = record.account.take() {
+					AccountToPersonalId::<T>::remove(&account);
+					frame_system::Pallet::<T>::dec_sufficients(&account);
```
gui1117, line 977: I wonder if we should write associated properties directly in the doc with some clear label like: `invariant: the field account has a related sufficient reference that should be added when created and removed when destroyed`. So the AI and us can ensure those properties more easily.

We can also have it in the type system but it is often annoying to write.
Like for pallet-score there was a `credit` field with associated locked funds that I never released when the user offboarded if he didn't redeem before.

gui1117 (approved):

andrzejSulkowski (approved):

pallets/people-multi/src/lib.rs
georgepisaltu: Added.

PR #51, opened 2026-08-24
dotns-gateway: add AccountNames reverse map

### LINKED ISSUE #49: dotns-gateway: subscribable accountId → registered-names mapping for app integration
## Context

Usernames moved from People Chain `Resources.Consumers` to dotNS on Asset Hub (individuality#755, host-rust-core#426). Apps previously subscribed to `Resources.Consumers` to resolve and live-update usernames (own account + contact lists). The Polkadot App is migrating now (polkadot-app-android-v2 PRs 978/992, dotns#243).

On Asset Hub today there is no chain-native way to subscribe to "what names does this accountId hold":

- `AccountAlias` (accountId → alias) is written exactly once, at the first full registration inside `register_name` — a one-shot edge trigger. It never updates afterwards and covers nothing about lite labels or later names.
- `LiteLabelOwner` is keyed label → owner: the wrong direction, not enumerable per account.
- Contract state (`DotnsRegistrar`/`LabelStore`) holds the truth, but the AH nodes expose no `eth_*` RPC (verified via `rpc_methods` on both public nodes), so `eth_getLogs`/log subscriptions are unavailable without an eth-rpc adapter deployment.
- `System.Events` scanning works (both lite and full registrations emit events), but forces every client to decode every block — a poor fit for mobile (battery/bandwidth).

Full analysis with the app-side requirements: dotns#243 (comment with Branch A/B comparison).

## Proposal

Add a reverse mapping to `pallet_dotns_gateway`, maintained on the pallet-mediated registration paths:

- `AccountNames: AccountId → BoundedVec<(Label, NameClass), MaxNamesPerAccount>` — or, if we want the minimal shape, accountId → current primary full name + node.
- Written in `register_name` (full names) and on lite auto-link, i.e. everywhere the pallet already writes `LiteLabelOwner`/`AccountAlias`.
- Storing the **node** (namehash) alongside the label lets clients derive everything else via existing views (e.g. `DotnsPopResolver.chatKey(node)`) without a second pallet field.

This gives apps a single multi-key `state_subscribeStorage` over their contact list — the same integration shape they had with `Resources.Consumers`.

## Known limitation (by design)

The pallet only observes its own dispatch paths. Name transfers/expiry that happen purely contract-side (ERC721 transfer, escrow) would not update the map. Options, in order of preference:

1. Accept staleness for v1 (transfers are rare; apps re-verify on read via view calls) and document it.
2. Have the gateway contracts call back into the pallet on transfer (larger change, touches contracts — out of scope for the runtime window).

## Backfill

Registrations that already happened need a one-off backfill (enumerate `LabelStore` pages / existing `AccountAlias` entries in a migration), or the map starts empty and only tracks new registrations, with apps falling back to view-call resolution for pre-existing names. Preference: start empty + lazy touch-up on next pallet interaction, to keep the migration trivial for the fellowship-runtime deadline.

## Minimal v1 scope (what the PR will contain)

To fit the fellowship-runtime window, the first PR is intentionally minimal:

- The `AccountNames` storage map, written on the pallet's existing registration paths (`register_name` + lite auto-link) — no new dispatchables.
- Starts empty: no backfill migration; apps fall back to view-call resolution for pre-existing names.
- Transfer staleness accepted and documented (option 1 above).

## Timing

Fellowship runtimes are planned to merge end of this week; if this lands in time it rides that release, otherwise the next one. The app migration (PR 978) can ship against `System.Events` scanning in the interim and switch to the map when live.

I'm working on this — PR to follow shortly.

## Alternatives considered

- **Contract-lens route** (dotns#216 → #217 `profileOf(address)`): solves point-reads, not subscriptions; #216 is on hold pending #212. Complementary, not competing — see dotns#243.
- **eth-rpc adapter deployment** for `eth_getLogs`: infra dependency, no light-client story.
- **`System.Events` scanning**: works today, kept as the fallback; this issue is about removing the per-block cost.

filip-parity:
Closes #49

Apps migrating off People Chain `Resources.Consumers` (dotns#243, polkadot-app-android-v2#978) need a chain-native way to resolve and watch "what names does this account hold" on Asset Hub. The existing storage does not provide it: `AccountAlias` is written once at first full registration, `LiteLabelOwner` is keyed label → owner, and the public nodes expose no `eth_*` RPC for contract log subscriptions.

This PR adds the minimal reverse map. `AccountNames` stores an `AccountNameRecord { lite, full }` per account, written on the two registration paths the pallet already mediates (`reserve_name` and `register_name`). One storage key per account gives clients a multi-key `state_subscribeStorage` over a contact list, the same integration shape `Resources.Consumers` had.

Scope decisions (details in #49):

- Starts empty, no backfill migration; clients fall back to view-call resolution for pre-existing names.
- Labels that change purely contract-side (for example transfers) are not reflected; documented on the storage item.
- No new `Config` items, events or dispatchables, so no runtime wiring is needed.

## Changes

- `pallet-dotns-gateway`: new `AccountNames` storage map (`AccountId → AccountNameRecord`), populated in `reserve_name` (lite label, latest wins) and `register_name` (full label).
- `types`: new `AccountNameRecord` with optional `lite` and `full` labels.
- Tests: record written on both paths, later reservation overwrites the lite label, each write preserves the other field; existing `assert_noop` failure tests cover the no-write error paths.
- Benchmarks: both verify blocks now assert the record exists. Weights are stale (each extrinsic gained one read and one write); needs `/cmd bench --pallet indiv_pallet_dotns_gateway` before merge to regenerate the pallet and runtime copies.

Agreed with @Zebedeusz in https://github.com/paritytech/individuality-community/issues/49#issuecomment-5395416457 that this map is temporary: the long-term solution is direct contract-state observation, tracked in https://github.com/paritytech/individuality-community/issues/52.


pallets/dotns-gateway/src/lib.rs
```diff
@@ -182,6 +182,16 @@ pub mod pallet {
 	pub type LiteLabelOwner<T: Config> =
 		StorageMap<_, Blake2_128Concat, BaseLabel, T::AccountId, OptionQuery>;
 
+	/// The dotNS labels each account acquired through this gateway.
+	///
+	/// Written on [`Pallet::reserve_name`] and [`Pallet::register_name`] only, so
```
Zebedeusz: Underline the limitation more clearly in the comment, so that it's well visible that it's important.
Also note that this should be migrated to apps observing contract storage instead.

pallets/dotns-gateway/src/types.rs
```diff
@@ -126,6 +126,30 @@ impl From<[u8; 65]> for ChatKey {
...
+	Debug,
+	Default,
+	Encode,
+	Decode,
+	MaxEncodedLen,
+	TypeInfo,
+	DecodeWithMemTracking,
+)]
+pub struct AccountNameRecord {
+	/// The lite-person label (`<dns-stem>.<digits>`), if one was reserved.
```
Zebedeusz: Reserved is only full-person username.
Lite-username is always registered.

pallets/dotns-gateway/src/lib.rs
Zebedeusz, line 391: If they also reserve a full-person name, no need to store it too?
Maybe not, just asking.

pallets/dotns-gateway/src/lib.rs
filip-parity: Done in 108ea6f — the doc now leads with a warning that the contracts are the source of truth, spells out that only the pallet's own paths are reflected (transfers are not, clients re-verify via views) and links https://github.com/paritytech/individuality-community/issues/52 for the migration.

pallets/dotns-gateway/src/types.rs
filip-parity: Right, fixed in 108ea6f — it now says "most recently registered lite-person label".

pallets/dotns-gateway/src/lib.rs
filip-parity: I'd leave it out: the reserved base name isn't a live username yet — it only becomes one when it's claimed through register_name, which then writes `full`. It also lapses after the reservation window, and the pallet can't observe the lapse, so the map entry could go stale in a way clients can't detect. Anything that needs pending-claim state can read the contract's pendingClaims view.

pallets/dotns-gateway/src/types.rs
Zebedeusz: Good but fix it here as-well

filip-parity: Addressed, 4fb54f9

Zebedeusz (approved):

pallets/dotns-gateway/src/lib.rs
```diff
@@ -182,6 +182,21 @@ pub mod pallet {
...
 
+	/// The dotNS labels each account acquired through this gateway.
+	///
+	/// Keyed by account so clients can watch a set of accounts with one storage
+	/// subscription.
+	///
+	/// Warning: this map is not the source of truth; the dotNS contracts are. It
+	/// is written on [`Pallet::reserve_name`] and [`Pallet::register_name`] only,
+	/// so a label that changes purely contract-side (for example a transfer) is
+	/// not reflected here and clients must re-verify on read via contract views.
```
re-gius, line 193: Please add _"Names registered outside these paths never appear here at all, including pre-launch whitelist allocations, which claimants register directly against the dotNS registrar controller."_ or an equivalent sentence

re-gius (approved):
LGTM - just one comment addition

Issue #56, opened 2026-08-24
Bookkeeping: Release Notes


kianenigma:
So far, most of this work has been deployed on private testnets. In these environments we didn't bother writing runtime migrations (any implementation of `OnRuntimeUpgrade`, either as a standalone script or as pallet `Hooks`).

We need to move away from this paradigm and organize each PR better. My suggestion on how to go forward is as follows: 

The end outcome we want is for each release (including the new nightly releases by @mordamax #14) note, we should be able easily pass the release link with the WASM artifacts to an external team, being it external or parity's SRE team, and expect them to be able to upgrade a (test)network to this release. There is a lot of information that eventually might be needed for this, but the most basic ones are: 

* Is this WASM release compatible with the previous one, or does it require a wipe 
  * Ideally, this should never happen again
* Does this WASM release require a new `RuntimeUpgrade` implementation to be added at the runtime level? 
  * In most cases, we should handle this with `Hooks`.
  * Also we should start using `pallet_version` properly for this. With the right pallet version, the CI of a downstream project wanting to integrate these runtimes might catch them if they use `try-runtime-cli`. 
* Does this this WASM release require a node upgrade? 
  * Rare case. Example is use of a new host function by a pallet
* Does this PR require the new `trait Config` items to be integrated in the runtime level  

To represent this, I suggest we tag each PR with the following labels (names TBD -- `C` for compatibility): 
* `C0-non-breaking`: This PR is just patch. A release containing it should be safely updated. Most PRs should be like this. A new pallet that is just added can also be like this. 
* `C1-runtime-migration`: This PR contains a runtime upgrade.
* `C2-runtime-config`: This PR alters the `type` in the `Config` trait of the pallet and requires code changes to downstream runtimes
* `C3-breaking-node`: This PR requires nodes to be upgraded to a specific version. 

This is really the basic building blocks of attaching some information to each PR, and represent them in our release notes. A lot more can be built here. Look at how complicated the release process of polkadot-sdk, with PRDocs and semver and so on, has gotten. We don't want to go that far, this is a tiny step in that direction.

The most common scenario to keep in mind here is: 

Suppose a deployed version of our `paseo-next` runtimes is at release `1.0.0` of this repo. All release of this bump major version (no semver)
Since then, there has been:
* `2.0.0` with at least one PR containing `C1-runtime-migration`
* `3.0.0` with at least one PR containing `C1-runtime-migration`
[...]

PR #58, opened 2026-08-25
Pallet-origin-restriction: use relay chain block number

gui1117:
Parachain block number frequency is variable, it is better to use time or relay chain block number.

I would prefer to use time, but given it is already used in production, using relay chain block number is less of a breaking change.


pallets/origin-restriction/src/lib.rs
```diff
@@ -115,7 +116,8 @@ pub mod pallet {
 	pub struct Usage<Balance, BlockNumber> {
 		/// The amount of usage consumed at block `at_block`.
 		pub used: Balance,
-		/// The block number at which the usage was last updated.
+		/// The block number at which the usage was last updated. It comes from
+		/// [`Config::BlockNumberProvider`], not from `frame_system`.
```
georgepisaltu: Maybe verbose.

georgepisaltu (approved):

pallets/origin-restriction/src/lib.rs
gui1117:
```suggestion
		/// The block number at which the usage was last updated. (Provided by
		/// [`Config::BlockNumberProvider`]).
```

gui1117: yes, I rewrote a bit, still a bit verbose but probably ok.

kianenigma (changes requested):
hmmmm, this is a good example of https://github.com/paritytech/individuality-community/issues/56 🙈

As in, I presume this block number, which was previously a para, and now is RC, is stored in the state, and I guess swapping it will have consequences. 

As in, we either need to wipe the state to not have corrupt data in this pallet, or we need to have a runtime migration script that will deal with it. 

Maybe I am wrong, but requesting change to discuss it more. 

Ignore my review if this is critical for safetynet. If not critical, let's wait on it a bit more.

gui1117:
I was hoping to include it to avoid a later migration, but ok let's hold off for now.

kianenigma (approved):
Okay thanks for acknowledging, your point actually is better than mine, that we will soon upgrade both paseo and polkadot with this, from the genesis mode, so there is real value in merging this. 

I am not sure about ppn, but I guess we will deal with that later, and it can be wiped again if needed.

PR #60, opened 2026-08-25
dotns-gateway: add chat key to AccountNames and backfill lite labels

filip-parity:
#### TL;DR

Follow-up to https://github.com/paritytech/individuality-community/pull/51 (#49), from the app team's feedback on https://github.com/paritytech/dotns/issues/243.

Two gaps in the `AccountNames` map as merged:

- The app needs the chat key on the hot path (verifying an incoming chat request). Reading it from the resolver costs two contract calls per peer with no subscription, and the pallet already receives the key in both registration paths, so storing it next to the labels is nearly free.
- The map starts empty, so accounts that registered before the upgrade would show no name once the app stops reading the People Chain. The pallet's own `LiteLabelOwner` map holds every lite label ever registered through the gateway, so it can be inverted in a migration. Live state today: 75 lite labels on paseo-next Asset Hub, 23 on previewnet, no full registrations on either, and no mainnet deployment yet, so lite labels are all there is to backfill.

#### Changes

- `AccountNameRecord` now holds a `NameEntry { label, chat }` per label, matching the contracts, which keep one chat key per label. `reserve_name` records the lite key; `register_name` records the given key for `Link::None`, and for `Link::LiteUsername` copies the recorded lite entry's key when the linked label is that entry (otherwise `None`).
- New `migration::MigrateV0ToV1` (storage version 0 → 1): translates any record stored in the two-field shape from #51, then fills `AccountNames` from `LiteLabelOwner`, keeping existing entries. Backfilled labels have `chat: None`, the pallet never stored their key. Wired into `migrations::Unreleased` of `next-asset-hub-paseo`.
- Tests: chat key on both write paths, inheritance on lite-linked registration, `None` when linking to a label other than the recorded one, standalone registration leaving the lite key alone; migration translates old records, backfills, keeps an existing lite entry, and runs only from version 0.


pallets/dotns-gateway/src/migration.rs
Zebedeusz, line 1: Most probably not needed since testnets have usually been purged before such updates and that does not exist on production network yet.
Check with @csmnprschv

pallets/dotns-gateway/src/types.rs
```diff
@@ -126,6 +126,19 @@ impl From<[u8; 65]> for ChatKey {
...
+#[derive(
+	Clone, PartialEq, Eq, Debug, Encode, Decode, MaxEncodedLen, TypeInfo, DecodeWithMemTracking,
+)]
+pub struct NameEntry {
+	/// The label.
+	pub label: BaseLabel,
+	/// The ECDH chat key the contracts hold for this label. `None` when the
+	/// pallet never saw the key: a label backfilled by migration, or a full
+	/// label linked to a lite label other than the recorded one.
+	pub chat: Option<ChatKey>,
```
Zebedeusz, line 139: Shouldn't the chat key be per name record? If I'm not mistaken the user may only use one chat key at a time.

pallets/dotns-gateway/src/tests.rs
```diff
@@ -328,7 +344,10 @@ mod reservation {
 
 			assert_eq!(
 				AccountNames::<Test>::get(ALICE),
-				Some(AccountNameRecord { lite: Some(base_name(BOB_LITE)), full: None })
+				Some(AccountNameRecord {
+					lite: Some(entry(BOB_LITE, Some(default_chat_key()))),
```
Zebedeusz: Since you always unwrap the value with `Some`, it'd be easier to just return an `Option` in fn entry

pallets/dotns-gateway/src/tests.rs
```diff
@@ -840,6 +868,67 @@ mod registration {
 		});
 	}
 
+	#[test]
+	fn lite_link_to_another_label_leaves_chat_key_unknown() {
+		new_test_ext().execute_with(|| {
+			System::set_block_number(1);
+			seed_lite_owner(ALICE_LITE, ALICE);
```
Zebedeusz: By the way, I just noticed that this fn is just a one-line storage insert so I suggest unwrapping it. It's clearer to read.

pallets/dotns-gateway/src/tests.rs
```diff
@@ -840,6 +868,67 @@ mod registration {
 		});
 	}
 
+	#[test]
+	fn lite_link_to_another_label_leaves_chat_key_unknown() {
```
Zebedeusz: How come "chat key unknown"? It's known - it's in the lite entry isn't it?

pallets/dotns-gateway/src/tests.rs
```diff
@@ -840,6 +868,67 @@ mod registration {
...
+					lite: Some(entry(ALICE_LITE, Some(lite_chat_key()))),
+					full: None,
+				},
+			);
+
+			assert_ok!(DotnsGateway::register_name(
+				person_registration_origin(alias_a()),
+				ALICE,
+				base_name(ALICE_BASE),
+				Link::None(default_chat_key())
```
Zebedeusz, line 906: I see this use case as the user not wanting the link and you do this link regardless.
Why so?

pallets/dotns-gateway/src/migration.rs
```diff
@@ -0,0 +1,123 @@
...
+	fn entry_without_chat(label: BaseLabel) -> NameEntry {
+		NameEntry { label, chat: None }
+	}
+
+	/// Use [`super::MigrateV0ToV1`] rather than this directly.
+	///
+	/// Translates any record stored in the two-field shape, then fills [`AccountNames`] from
+	/// [`LiteLabelOwner`] for lite labels registered before the map existed. Only the label is
+	/// recoverable from pallet storage: the chat key stays `None` for these accounts. When an
+	/// account owns several lite labels, the first one in storage iteration order wins. Full
```
re-gius: This is in contrast to what happens in `reserve_name` where you overwrite `record.lite` on every reservation, so live behaviour is latest wins.
So, theoretically, two accounts with identical histories can end up showing different labels depending on whether they were backfilled or registered after the upgrade.
This is very rare, but I would still document the difference by stating that backfilled entries carry no ordering guarantee.

pallets/dotns-gateway/src/migration.rs
```diff
@@ -0,0 +1,123 @@
...
+			let owners = LiteLabelOwner::<T>::iter().count() as u32;
+			Ok((records, owners).encode())
+		}
+
+		#[cfg(feature = "try-runtime")]
+		fn post_upgrade(state: alloc::vec::Vec<u8>) -> Result<(), sp_runtime::TryRuntimeError> {
+			let (records, owners) = <(u32, u32)>::decode(&mut &state[..]).map_err(|_| {
+				sp_runtime::TryRuntimeError::Other("pre_upgrade state is not (u32, u32)")
+			})?;
+			ensure!(
```
re-gius, line 121: Why this is not strict equality? Are there any names for which decoding fails?

re-gius (commented):
Please also add the comment on whitelisted names that cannot be recovered

https://github.com/paritytech/individuality-community/pull/51/changes#r3844273018

pallets/dotns-gateway/src/migration.rs
filip-parity: paseo-next AH has 75 lite names, previewnet 23, no full names anywhere. If they're purged this is a no-op, if not the app loses those names. @csmnprschv is paseo-next purged with this upgrade?

pallets/dotns-gateway/src/types.rs
filip-parity: Contracts key it per node, so per label. A lite name and a standalone full name can hold different keys, and then one key per record points at the wrong name. Can flatten if you'd rather, but per label is the honest one.

pallets/dotns-gateway/src/tests.rs
filip-parity: It's used in 8 tests, so that's 8 copies of the insert...

Zebedeusz: It's 8 replacements of fn call with insert, isn't it?

pallets/dotns-gateway/src/types.rs
Zebedeusz: I see, and what use case is there to store two keys? In the contracts even. Won't it only be a cause of confusion?

filip-parity: None that I know of, in practice the user has one key. Two only exist if someone registers a full name standalone (Link::None) while holding a lite name — the contracts allow it because names are independent tokens?

Zebedeusz: It's worth checking with the App guys - maybe they have a use for lite-person key even if full-person key was provided. If not, I'd suggest modifying it to store ever only one in all spots (contracts, runtime). Fine for me to start with an issue for it and carry it on later.

re-gius (approved):
LGTM

assuming this migration will only be run under the current state of things

filip-parity:
> LGTM
> 
> assuming this migration will only be run under the current state of things

@Zebedeusz give this pull request some love pretty please

Zebedeusz:
> > LGTM
> > assuming this migration will only be run under the current state of things
> 
> @Zebedeusz give this pull request some love pretty please

I've been waiting for replies to my comments.

pallets/dotns-gateway/src/tests.rs
filip-parity: Nothing gets linked here. `Link::None` registers the full name with its own fresh key; the lite entry is left exactly as it was because the account still owns that name. The record holding both entries just means "this account has these two names" — the contracts hold both too.

pallets/dotns-gateway/src/types.rs
filip-parity: cc @foxwoosh - is a lite key ever needed once a full key exists? If not, the change is contract-side first and the runtime follows. Either way this should be tracked in another issue so this PR stays scoped to the backfill...

filip-parity:
> I've been waiting for replies to my comments.

Every comment has been either addressed (i.e. `outdated`) or responded to

pallets/dotns-gateway/src/tests.rs
Zebedeusz: What I meant is that the link to lite-user keeps existing and while it's easy for us to do this because we know that it's the same account that has some lite-person name entry, I think we should not keep it and either retain two entries somehow but with no link or only the full-person entry.
Fine for me to assess it in a separate issue.

pallets/dotns-gateway/src/migration.rs
Zebedeusz: I would still prefer checking that before merging - it would be cleaner to go to Polkadot without any migration since it will be the first arrival of the pallet there.

Zebedeusz (approved):
All looks fine but please check if we indeed need the migration - as per my comment

pallets/dotns-gateway/src/tests.rs
filip-parity: Addressed

pallets/dotns-gateway/src/tests.rs
filip-parity: Addressed

pallets/dotns-gateway/src/tests.rs
filip-parity: Addressed

Issue #6, opened 2026-08-18
[Members Subscriber] Move the ring gap scan to an offchain worker call


Zebedeusz:
`detect_missing_rings_in_batch` is called only from `initialize_ring_roots` and `process_ring_updates`, so gap detection runs only while the notifier is sending batches.

Problems that arise from that:
- detection stops when the notifier sends nothing,
- the notifier's XCM pays for subscriber bookkeeping,
- multi-part init records false gaps. Each part scans against a `next_ring_index` the later parts have not provided yet, so indices on-the-way are recorded as missing. 

**Solution**
An authorized call, submitted by the offchain worker, that scans one page and records gaps. 
Then remove the scan from both notifier extrinsics.

PR #61, opened 2026-08-25
[Members subscriber] missing indices scan moved to offchain worker

Zebedeusz:
Closes #https://github.com/paritytech/individuality-community/issues/6

Before this change, missing indices detection used to run in `initialize_ring_roots` and `process_ring_updates` so only when ring roots were received from the notifier.
The proposed solution is to move this effort to offchain worker in order to detach this logic from updates receival. 

## Changes

- Gap scan moved out of the batch extrinsics into a new authorized call `detect_missing_rings`, submitted by the offchain worker. `initialize_ring_roots` and `process_ring_updates` no longer pay for this scanning.
- `authorize_detect_missing_rings` validates the source, the subscription state, the scan cooldown and the collection's capacity.
- The offchain worker submits one scan per run and skips a collection that cannot make progress.
- Renamed `MaxGapScanPerBatch` to `MaxGapScanPerCall`. Added `GapScanCooldownSeconds`, the rest time after a batch before a scan runs.
- Offchain worker transactions carry a retry discriminator.
- `integrity_test` checks the new call against the block budget and the new constants against each other.
- New benchmarks for `detect_missing_rings` and its `authorize`, with a worst-case state that fills both index sets.


pallets/members-subscriber/src/lib.rs
```diff
@@ -73,13 +76,24 @@ pub mod pallet {
...
 	};
+	use sp_runtime::SaturatedConversion;
 	use xcm::v5::{Location, SendXcm};
 
 	const LOG_TARGET: &str = "pallet-members-subscriber";
 
 	/// Number of blocks that an offchain worker transaction of this pallet stays valid in the
 	/// transaction pool.
-	const TX_LONGEVITY: u64 = 3;
+	const TX_LONGEVITY: u64 = 8;
```
tindzk, line 87: Is this increase arbitrary, or related to the changes? It might be helpful explaining in the comment how this value was determined.

pallets/members-subscriber/src/lib.rs
```diff
@@ -441,6 +442,13 @@ pub mod pallet {
 				"OffchainWorkerInterval must be greater than 0"
 			);
 
+			// Retries deduplicate only when several offchain worker runs share one retry window.
+			// At or above the window every run carries a fresh discriminator.
+			assert!(
+				T::OffchainWorkerInterval::get() < TX_RETRY_WINDOW.into(),
+				"OffchainWorkerInterval must be less than TX_RETRY_WINDOW"
+			);
+
```
tindzk, line 455: Should we also add an assertion for `OldRootRetentionDuration`?

```rust
assert!(
    T::OldRootRetentionDuration::get() > 0,
    "OldRootRetentionDuration must be greater than 0"
);
```

Or more strictly, `OldRootRetentionDuration > 0 || MaxRecentRootsPerRing == 1`.

pallets/members-subscriber/src/lib.rs
```diff
@@ -649,14 +645,17 @@ pub mod pallet {
 		/// Submitted by the offchain worker as an authorized transaction. Validates
 		/// that the subscription is active and that the provided indices are actually
 		/// missing before sending XCM replay requests.
-		#[pallet::authorize(Pallet::<T>::authorize_replay_missing_roots)]
+		#[pallet::authorize(|source, identifier, indices, _discriminator| {
+			Pallet::<T>::authorize_replay_missing_roots(source, identifier, indices)
```
tindzk, line 653: `authorize_replay_missing_roots` should return the unspent weight rather than the full weight.

The same issue is present in `authorize_purge_stale_ring_roots`.

pallets/members-subscriber/src/benchmarking.rs
```diff
@@ -469,44 +500,51 @@ mod benches {
 		Ok(())
 	}
 
-	/// Benchmark for the scan loop in `detect_missing_rings_in_batch`.
-	/// Measures cost of scanning a range of `n` indices, swept over the full per-batch
-	/// scan cap since the extrinsics charge up to `MaxGapScanPerBatch` indices.
 	#[benchmark]
-	fn detect_missing_in_range(
-		n: Linear<1, { T::MaxGapScanPerBatch::get() }>,
+	fn detect_missing_rings(
```
tindzk, line 525: This doesn't seed `RingRoots`, so `present` will always be false:

```rust
let present = RingRoots::<T>::contains_key((generation, identifier, idx));
```

pallets/members-subscriber/src/tests.rs
```diff
@@ -2843,10 +3112,232 @@ mod scan_cap_and_weights {
...
+			let first = call.authorize(TransactionSource::InBlock).unwrap().unwrap().0;
+
+			System::set_block_number(2);
+			let retry = call.authorize(TransactionSource::InBlock).unwrap().unwrap().0;
+
+			// Sharing a tag, the retry is only accepted by the pool if it strictly outbids.
+			assert_eq!(first.provides, retry.provides);
+			assert!(retry.priority > first.priority);
+		});
+	}
```
tindzk, line 3337: We can make the test more specific:

```suggestion
	#[test]
	fn gap_scan_priority_rises_with_the_block_it_is_validated_in() {
		new_test_ext().execute_with(|| {
			setup_lagging_scan(100);
			let call = gap_scan_call(PEOPLE);

			System::set_block_number(1);
			let early = call.authorize(TransactionSource::InBlock).unwrap().unwrap().0;

			System::set_block_number(2);
			let late = call.authorize(TransactionSource::InBlock).unwrap().unwrap().0;

			// Retries share the stored-cursor tag, so the pool ranks them by priority alone.
			assert_eq!(early.priority, tx_priority::BACKGROUND_PROGRESS + 1);
			assert_eq!(late.priority, tx_priority::BACKGROUND_PROGRESS + 2);
		});
	}
```

pallets/members-subscriber/src/lib.rs
Zebedeusz: Added an explanation.

pallets/members-subscriber/src/lib.rs
Zebedeusz: `OldRootRetentionDuration` == 0 means no grace period, so only each ring's newest root verifies. 
That is a valid and the safest setting, so I would rather not prohibit it.

pallets/members-subscriber/src/lib.rs
Zebedeusz: Fixed

pallets/members-subscriber/src/benchmarking.rs
Zebedeusz: Added

pallets/members-subscriber/src/tests.rs
Zebedeusz: Done

tindzk (approved):

andrzejSulkowski:
Time to take care of this 3-week-old beauty

pallets/members-subscriber/src/lib.rs
```diff
@@ -800,9 +831,64 @@ pub mod pallet {
...
+			if !matches!(source, TransactionSource::InBlock | TransactionSource::Local) {
+				return Err(InvalidTransaction::Call.into());
+			}
+
+			if !matches!(Subscription::<T>::get(), SubscriptionStatus::Active { .. }) {
+				return Err(InvalidTransaction::Call.into());
+			}
+
+			let now = T::UnixTime::now().as_secs();
+			let last_batch = ProcessingState::<T>::get().last_batch_received_time;
```
andrzejSulkowski, line 858: following scenario which leads to a starving collection:

Collection A receives batches more often than every 60s. Collection B has been quiet for hours with a real gap. B's scan never becomes eligible: `authorize_detect_missing_rings` reads the global `last_batch_received_time`, sees it fresh because of A, and returns Future.

runtimes/next-asset-hub-paseo/src/weights/indiv_pallet_members_subscriber.rs
```diff
@@ -308,4 +308,21 @@ impl<T: frame_system::Config> indiv_pallet_members_subscriber::WeightInfo for We
...
 	}
+	/// Storage: `MembersSubscriber::Subscription` (r:1 w:0)
+	/// Proof: `MembersSubscriber::Subscription` (`max_values`: Some(1), `max_size`: Some(9), added: 504, mode: `MaxEncodedLen`)
+	/// Storage: `Timestamp::Now` (r:1 w:0)
+	/// Proof: `Timestamp::Now` (`max_values`: Some(1), `max_size`: Some(8), added: 503, mode: `MaxEncodedLen`)
+	/// Storage: `MembersSubscriber::ProcessingState` (r:1 w:0)
+	/// Proof: `MembersSubscriber::ProcessingState` (`max_values`: Some(1), `max_size`: Some(24), added: 519, mode: `MaxEncodedLen`)
+	/// Storage: `MembersSubscriber::RingCollectionStates` (r:1 w:0)
+	/// Proof: `MembersSubscriber::RingCollectionStates` (`max_values`: None, `max_size`: Some(2504), added: 4979, mode: `MaxEncodedLen`)
+	fn authorize_detect_missing_rings() -> Weight {
```
andrzejSulkowski, line 319: - [ ] check and close after weight regen

pallets/members-subscriber/src/lib.rs
```diff
@@ -1133,20 +1215,129 @@ pub mod pallet {
...
-				Ok(()) => log::debug!(
-					target: LOG_TARGET,
-					"offchain worker: submitted authorized transaction for `{description}`"
-				),
-				Err(()) => log::warn!(
-					target: LOG_TARGET,
-					"offchain worker: failed to submit authorized transaction for \
-					 `{description}`"
-				),
+			if SubmitTransaction::<T, _>::submit_transaction(tx).is_ok() {
```
andrzejSulkowski, line 1333: `submit_authorized_transaction` is being used by various methods. I think its not good that we now only log on stalls. Like failed ring purge tx's would go now unnoticed.

pallets/members-subscriber/src/lib.rs
```diff
@@ -1133,20 +1215,129 @@ pub mod pallet {
...
+				return;
+			}
+
+			for (identifier, state) in RingCollectionStates::<T>::iter() {
+				if state.missing_indices.is_empty() {
+					continue;
+				}
+
+				if state.deleted_indices.len() as u32 >= T::MaxDeletedRingsPerCollection::get() {
+					Self::warn_periodically(
```
andrzejSulkowski, line 1302: `fn warn_periodically` docs say: "Logs a warning once per STALL_WARN_PERIOD blocks, and a debug message otherwise.". Idk if this makes here really sense. As soon as `deleted_indices >= T::MaxDeletedRingsPerCollection::get()`, we want to immediately emit a warning or even an error - I guess.

pallets/members-subscriber/src/lib.rs
```diff
@@ -1133,20 +1215,129 @@ pub mod pallet {
...
+				generation: progress.generation,
+				page: progress.page,
+				discriminator,
+			};
+			Self::submit_authorized_transaction(call, block_number);
+		}
+
+		/// Submits one gap-scan transaction for the first collection whose scan cursor is behind
+		/// the ring index frontier.
+		fn submit_gap_scan(
```
andrzejSulkowski, line 1234: nit: not a big fan of the fn names. I thought initially it literally submit the tx, but they perform a bunch of checks first and only maybe submit.

pallets/members-subscriber/src/lib.rs
```diff
@@ -954,82 +1041,77 @@ pub mod pallet {
...
+		/// The scan resumes at `next_scan_index` and covers at most `MaxGapScanPerCall` indices.
 		/// Returns the number of indices examined, for the caller's weight refund.
-		pub(crate) fn detect_missing_rings_in_batch(batch: &RingRootUpdatesBatch<T>) -> u32 {
-			let state = RingCollectionStates::<T>::get(batch.identifier);
+		pub(crate) fn detect_missing_rings_in_collection(identifier: Identifier) -> u32 {
+			let state = RingCollectionStates::<T>::get(identifier);
 
 			let accounted_for = state.ring_count.saturating_add(state.deleted_indices.len() as u32);
 
 			if accounted_for >= state.next_ring_index {
```
andrzejSulkowski, line 1066: I think we might have a pre-existing issue. What if we hit the `ReplayAbandonThreshold`, lets say with 1 idx and hit this line of code. We end up with `accounted_for` at most `next_ring_index - 1`. This might cause more issues in the code we are not yet aware of.

pallets/members-subscriber/src/lib.rs
```diff
@@ -800,9 +831,64 @@ pub mod pallet {
...
+			if !matches!(Subscription::<T>::get(), SubscriptionStatus::Active { .. }) {
+				return Err(InvalidTransaction::Call.into());
+			}
+
+			let now = T::UnixTime::now().as_secs();
+			let last_batch = ProcessingState::<T>::get().last_batch_received_time;
+			if now.saturating_sub(last_batch) < T::GapScanCooldownSeconds::get() {
+				// The cooldown elapses on its own, so the pool keeps the transaction and
+				// revalidates it in a later view.
+				return Err(InvalidTransaction::Future.into());
```
andrzejSulkowski, line 862: If I am not mistaken, `Err(Future)` parks the transaction for 30 min into the pool and only after that time it will be re-validated. During this time the same transaction can not be re-submitted because of duplicates.
Now if `state.next_scan_index` doesnt move, we end up in a 30 min deadlock.

We could prevent this by taking `BlockNumber / TX_LONGEVITY` into the `and_provides()` and making every tx unique across `TX_LONGEVITY` windows

pallets/members-subscriber/src/lib.rs
andrzejSulkowski: or lets put it that way: how severe is it, if deleted_indices get saturated?

pallets/members-subscriber/src/lib.rs
```diff
@@ -1133,20 +1215,129 @@ pub mod pallet {
...
+			block_number: BlockNumberFor<T>,
+			discriminator: BlockNumberFor<T>,
+			now: u64,
+			processing_state: &UpdatesProcessingState,
+		) {
+			// Some rest time after the last received batch is given to account
+			// for incoming batches that may contain indices considered as missing
+			// in the current state.
+			if now.saturating_sub(processing_state.last_batch_received_time)
+				< T::GapScanCooldownSeconds::get()
```
andrzejSulkowski, line 1244: Busy collections block all gap detection. I dont know the solution for that, but its definitely not great as soon as we get traction that this will not or barely get executed.

pallets/members-subscriber/src/lib.rs
```diff
@@ -1133,20 +1215,129 @@ pub mod pallet {
...
+					continue;
+				}
+
+				let call = Call::detect_missing_rings {
+					identifier,
+					scan_from: state.next_scan_index,
+					discriminator,
+				};
+				Self::submit_authorized_transaction(call, block_number);
+				return;
```
andrzejSulkowski, line 1273: I think this return statement makes more sense *after* the for-loop.

pallets/members-subscriber/src/benchmarking.rs
andrzejSulkowski, line 103: a side finding.
`lib.rs:794` does `let has_missing = indices.iter().any(|idx| state.missing_indices.contains_key(idx));`.

If we now have our missing starting from 0, it will land a hit on the first lookup.

andrzejSulkowski, line 228: side finding: we could probably get here `worst_case_collection_state` to get the most PoV size state. Right now deleted_indices are for example set to 0.

andrzejSulkowski (changes requested):
few comments above which I'd classify as blocking until resolved.

andrzejSulkowski:
And I'd suggest labeling this as a `needs-audit`

pallets/members-subscriber/src/lib.rs
Zebedeusz: As to the immediate log and event - this is already the case: pallets/members-subscriber/src/lib.rs:1057, pallets/members-subscriber/src/lib.rs:1039.
As to severity: it is severe because when at capacity it basically disables the gap detection.

pallets/members-subscriber/src/lib.rs
Zebedeusz: Isn't the transaction "park" done by id'ing the transactions by their hash? 
If so, the hash will change every `TX_RETRY_WINDOW` so every 8 blocks.

pallets/members-subscriber/src/lib.rs
Zebedeusz: It's on purpose this way so that at most one `detect_missing_rings` call is submitted.
This way we know it's cheap enough to nicely enter a block and it's more predictable and easier to benchmark.

pallets/members-subscriber/src/benchmarking.rs
Zebedeusz: These lines you mentioned are of `authorize_replay_missing_roots` and it has its own, different worst state setup.

Zebedeusz: I think that with `MaxEncodedLen` on `RingCollectionState`, the proof size is charged at the type's maximum regardless of content.

pallets/members-subscriber/src/lib.rs
Zebedeusz: I think it's fine because e.g. the `submit_ring_purge` is retried every time the offchain worker runs.
But I will add a certain tiny fix to it to make the logs more predictable.

pallets/members-subscriber/src/lib.rs
Zebedeusz: So in such a case, that index would be added to the missing indices, the cursor catches up with next_ring_index, then during each next authorize_detect_missing_rings an error NothingToScan would be returned, until that missing index becomes too old and is removed or received from the notifier (replays for it would be sent).
I don't really see an issue here.

pallets/members-subscriber/src/lib.rs
Zebedeusz: Should be fixed now.

pallets/members-subscriber/src/lib.rs
Zebedeusz: Renamed them.

pallets/members-subscriber/src/lib.rs
Zebedeusz: Should be resolved with the modifications related to your first comment.

pallets/members-subscriber/src/lib.rs
andrzejSulkowski: If we have multiple collections, some of them later in line would need to wait multiple of the waiting period to get their gaps closed. Furthermore, I dont see how this relates benchmarks at this is executed by a OCW?

PR #62, opened 2026-08-26
[Example]: Example and Doc on how coinage instances work and are created

andrzejSulkowski:
Core of this work is the doc `e2e/coinage-instance/Coinage Instance Example.md`, it explains how to create coinage instances and what the differences between them are.

This doc is accompanied by code which builds the actual calls and provides a working example.


agustinustheo:
When you're back, can you tell me if this needs a review?

There's ".bin" files and I don't know what those are for.

e2e/coinage-instance/chunks/r2e10/page-0.bin
agustinustheo, line 1: Nit: this is a comment for the entire directory not just this file, but initial look gets me a bit confused. Like why is this it's own directory?

I'll try to recommend something once I try to understand this better (or not if it's good)

andrzejSulkowski: these are the chunk pages which need to get send in multiple transactions. We had this data inline in our init scripts. 
The initial idea was to show external folks how coinage works so they can create their own instances and init them properly.

PR #66, opened 2026-08-26
Stamp per-nightly spec_version in runtime builds

mordamax:
## Changes

- `build-runtimes.yml`: new `nightly` input. When set, a step before the srtool build stamps `spec_version` as `committed base * 100 + count of nightly tags` (e.g. `2_000_039` → `200_003_904`), so every nightly strictly increases and a chain accepts it over any earlier nightly. The stamp lives in the build workspace only; nothing is committed back. The step hard-fails if the sequence reaches 100 (two spare digits), which a base never reaches because every manual release bumps it.
- `nightly-release.yml`: passes `nightly: true`; the tag gains a `-HHMM` suffix when the plain `nightly-YYYY-MM-DD` tag already exists, so a manual rerun in the same day creates its own release and bumps the sequence instead of overwriting the earlier release's assets with a same-version build.
- Release notes (both `nightly-release.yml` and `publish.yml`) print each runtime's `spec_version` prominently, read from the built wasm's own srtool output, so what is shown is guaranteed to be what shipped.

Example stamps for the coming days (asset-hub / people):

| Nightly | asset-hub | people |
|---|---|---|
| 2026-08-26 | 200_003_905 | 100_003_605 |
| 2026-08-27 | 200_003_906 | 100_003_606 |
| 2026-08-28 | 200_003_907 | 100_003_607 |

Caveat, documented in the step comment: a manual release that bumps the committed base by +1 lands below the nightly band (`2_000_040` < `200_003_905`), so a chain that applied a nightly rejects such a release. The release must bump past the stamped band, or a follow-up can extend stamping to `publish.yml`.

### DIFF
--- a/.github/workflows/build-runtimes.yml
+++ b/.github/workflows/build-runtimes.yml
@@ -1,9 +1,16 @@
 name: Build runtimes
 
 # Shared srtool build for the two paseo runtimes. Called by the manual Release
-# and the nightly pre-release workflows so both ship identical artifacts.
+# and the nightly pre-release workflows; the nightly passes nightly: true so its
+# builds get a timestamped spec_version.
 on:
   workflow_call:
+    inputs:
+      nightly:
+        description: Stamp spec_version with base * 100 + nightly sequence so each nightly upgrades
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   build-runtimes:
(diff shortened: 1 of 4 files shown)


.github/workflows/build-runtimes.yml
```diff
@@ -21,6 +28,38 @@ jobs:
...
+      # built from the raw source can never upgrade a chain already running a
+      # previous nightly. Stamp spec_version as base * 100 + N, where base is
+      # the committed value and N the count of earlier nightly tags: the
+      # stamped number starts with the committed digits (2_000_039 becomes
+      # 200_003_9NN) and strictly increases per nightly. spec_version is a
+      # u32, and base digits + N cap at 9 digits, which fits; N saturates at
+      # 99, so a base must not see 100 nightlies — every manual release bumps
+      # the base, which resets N. Manual releases must keep the committed
+      # value below the stamped band of any nightly already applied on-chain
+      # (bumping 2_000_039 to 2_000_040 stays under 200_003_900, so a chain
```
andrzejSulkowski, line 42: would it make sense in this case to establish the rule that the last two digits are reserved for nightly releases only?

```
200_000_000, casual bump -> 200_000_100
200_000_100, nightly release -> 200_000_101
```

mordamax: well, i wanted to preserve the original spec version for nightlies
base is 2_000_039
2_000_039 → 200_003_901
2_000_039 → 200_003_902
base upgraded
2_000_040 -> 200_003_901
2_000_040 -> 200_003_901

this way we can see what was the previous release that this nightly have diverged from, and it's clear also that next stable is 2_000_041

andrzejSulkowski: Right, but with this there is the issue you described here in the comment. A chain reject a nightly release because its greater the regular release. So it would be more handy with 2 extra digits reserved for nightly, I think? Beware, I am not a CI/CD guy :) I just want to share my thought here and learn.

What I can not estimate is how much trouble it would cause to deviate from the original nightlies spec version.

mordamax: I guess i didn't understand your initial suggestion, and now i finally realized

so you suggest to reestablish the baseline for both stable and unstable by adding 2 numbers? that'd work i think

like 

200_000_000, casual bump -> 200_000_100
200_000_100, nightly release -> 200_000_101
in 10 releases
200_000_900, casual bump -> 200_001_000
200_001_000, nightly bump -> 200_001_001
200_001_001, nightly bump -> 200_001_002
etc

kianenigma:
Will subsume this into https://github.com/paritytech/individuality-community/pull/74, don't review.

PR #67, opened 2026-08-26
Implement assets forwarder pallet

georgepisaltu:
This PR adds permissionless forwarding of assets from one chain to another (built for AH to PC trust backed asset registration). As in the paseo example in our repo, the assets being forwarded are force created and can only be minted via XCM transfers and are usable in pools. Force creation means there will be no issuer, so nobody can alter the issuance outside of XCM transfers, which are checked on AH. The owner of these assets will be the forwarder pallet on AH. An existing asset (like HOLLAR on PC) is unaffected as it was force created before this change.

Still TODO:
- review benchmarks
- compute weights
- zombienet test (though we have integration tests which look at the entire flow by examining the XCMP queue)

In another PR, we would need to port this to Polkadot runtimes.


pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,434 @@
...
+//! authenticates the program by this pallet's location, so its `pallet-assets` instance must set
+//! an origin filter (for example `EnsureXcm<Equals<location>>`) that matches
+//! `(1, [Parachain(<this chain>), PalletInstance(<this pallet>)])` as `ForceOrigin`.
+//!
+//! The replica's owner and team resolve to an account nobody controls on the destination, so its
+//! issuance there can only change through XCM transfers backed by this chain. The caller pays the
+//! XCM delivery fees and a deposit which backs the [`ForwardedAssets`] entry; the deposit is held
+//! forever, which is what makes junk forwards costly.
+//!
+//! Sufficiency and minimum balance can change after an asset was forwarded, so the permissionless
```
gui1117, line 35: what about metadatas? it can also be changed by the owner. Ideally metadata shouldn't exist in the foreign chain.

georgepisaltu: Indeed there is no need for metadata in this context, it can just be queried on the original chain by whoever builds apps over the people chain. I removed the metadata forwarding.

pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+		type Destination: Get<Location>;
+
+		/// Index of the `pallet-assets` instance in the destination's runtime call enum. The
+		/// destination decodes the remote calls under this index, so a mismatch makes every
+		/// forward fail on the destination.
+		type RemoteAssetsPalletIndex: Get<u8>;
+
+		/// Location of the local `pallet-assets` instance the forwarded assets live in. Asset ids
+		/// are appended to it as `GeneralIndex` junctions and reanchored to the destination.
+		type AssetsPalletLocation: Get<Location>;
```
gui1117, line 157: Maybe this can be deducted from the pallet assets. The index is available through the trait PalletInfo or PalletInfoAccess, then I believe it is standard to make the location of the asset 'Here, PalletIndex, GeneralIndex'.

On the other hand if we wanted to support the forward of foreign asset in this pallet then we would need to specify None as the location here I guess, so that forwarded asset like Eth would keep their same location on the destination (converted to the destination point of view).

It is just a nit, anyway the code will be fine.

pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+
+		/// Universal location of this chain, used to reanchor locations to the destination.
+		type UniversalLocation: Get<InteriorLocation>;
+
+		/// Converts this pallet's location, as seen from the destination, into the account that
+		/// owns the forwarded assets there.
+		type DestinationAccountOf: ConvertLocation<Self::AccountId>;
+
+		/// Converts an asset id into the `GeneralIndex` value of its location.
+		type AssetIdToIndex: Convert<Self::AssetId, u128>;
```
gui1117, line 167: Or you can change this to asset id to location and remove the associated type pallet assets location.

So it would support both local assets and foreign assets, foreign assets would convert to themselves.

(Also given Hollar is already created on PC, we would need to take it into account if we ever support foreign asset forwarding. We would need to do a migration to put the forwarded value in the state here on AH, or maybe let the force create fail on PC, anyway not an concern for now)

pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+			<T as Config<I>>::Currency::hold(
+				&HoldReason::<I>::ForwardDeposit.into(),
+				&who,
+				deposit,
+			)?;
+
+			let message = Self::build_remote_xcm(&create);
+			let message_id = Self::send_remote_xcm(origin, message)?;
+
+			ForwardedAssets::<T, I>::insert(
```
Zebedeusz, line 286: Here you assume all went well so I wonder if we don't need a mechanism to handle failures.
For example clearing the ForwardedAssets entry via an origin-gated call.

pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+			let asset_id: T::AssetId = id.into();
+			ensure!(
+				!ForwardedAssets::<T, I>::contains_key(&asset_id),
+				Error::<T, I>::AlreadyForwarded
+			);
+
+			let details =
+				pallet_assets::Asset::<T, I>::get(&asset_id).ok_or(Error::<T, I>::UnknownAsset)?;
+			ensure!(
+				details.status == pallet_assets::AssetStatus::Live,
```
Zebedeusz, line 263: Should we also check if `details.min_balance` is not zero?

pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+				ForwardedAssets::<T, I>::get(&asset_id).ok_or(Error::<T, I>::NotForwarded)?;
+
+			let details =
+				pallet_assets::Asset::<T, I>::get(&asset_id).ok_or(Error::<T, I>::UnknownAsset)?;
+			ensure!(
+				details.status == pallet_assets::AssetStatus::Live,
+				Error::<T, I>::AssetNotLive
+			);
+			ensure!(
+				record.min_balance != details.min_balance ||
```
Zebedeusz, line 325: If these assets can be created by anyone, then the owner could be changing these values and requesting the sync without limit.

pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+
+			let status = RemoteAssetsCall::ForceAssetStatus {
+				id: remote_asset_id,
+				owner: MultiAddress::Id(owner.clone()),
+				issuer: MultiAddress::Id(owner.clone()),
+				admin: MultiAddress::Id(owner.clone()),
+				freezer: MultiAddress::Id(owner),
+				min_balance: details.min_balance,
+				is_sufficient: details.is_sufficient,
+				is_frozen: false,
```
Zebedeusz, line 341: Can there be a scenario where the remote chain froze an asset on purpose?
In such a scenario this call would unfreeze it.

runtimes/next-people-paseo/src/integration_tests/asset_forwarding.rs
gui1117, line 1: if we introduce some test using the XCM emulator similarly to what exists on fellowship/runtimes we could have more accurate testing maybe. Anyway it is ok for now.

pallets/assets-forwarder/src/benchmarking.rs
```diff
@@ -0,0 +1,105 @@
...
+			2u32.into(),
+			true,
+			false,
+		)
+		.map_err(|_| BenchmarkError::Stop("failed to change asset status"))?;
+
+		#[extrinsic_call]
+		_(RawOrigin::Signed(caller), id);
+
+		Ok(())
```
gui1117, line 127: let's assert that the forwarded asset value is updated.

gui1117 (approved):
Some nits were opened: refactor of the config types and add an assertion to the benchmark. Otherwise all looks good.

pallets/assets-forwarder/src/lib.rs
gui1117: I think it should be fine. I don't see why we should limit what is on AH, if the asset is min balance 0, it is likely not sufficient, or maybe only distributed to some specific individuals.

pallets/assets-forwarder/src/lib.rs
gui1117: using TechnicalMaintenance instead of root would help indeed.

pallets/assets-forwarder/src/lib.rs
georgepisaltu: It will just fail on PC, no harm done and state on AH will be updated.

pallets/assets-forwarder/src/lib.rs
georgepisaltu: We should aim to mirror the state 1:1 with AH and this also lets us keep this pallet simple. I don't want to reinvent another `pallet-assets`.

pallets/assets-forwarder/src/lib.rs
georgepisaltu: Yes but they pay for the call. They can manually send it even without this particular call wrapper in this pallet by going straight to `pallet-xcm`. Obviously it will just fail on arrival and it will burn funds, so it's not a valid attack.

pallets/assets-forwarder/src/lib.rs
georgepisaltu: No, the remote chain isn't a "sovereign" 3rd party parachain, it's a system chain answering to the same governance as PC, therefore it does not need to assert control over the assets mirrored from AH.

pallets/assets-forwarder/src/benchmarking.rs
georgepisaltu: Done.

runtimes/next-people-paseo/src/integration_tests/asset_forwarding.rs
georgepisaltu: I'd defer this for later.

pallets/assets-forwarder/src/lib.rs
georgepisaltu: Added a call to remove the entry.

pallets/assets-forwarder/src/lib.rs
gui1117: the execution fee on the destination chain is waived, but it isn't much, the delivery fee are still charged at the moment.
We can probably ignore or charge a little more with a configurable constant.

pallets/assets-forwarder/src/lib.rs
gui1117: what would fail on PC? The Hollar forwarding? yes true, we don't even need migration. but this is not my main comment:

My proposition is to generalize this pallet by improving the `Config` trait:
* remove `AssetsPalletLocation` and `AssetIdToIndex`
* introduce `AssetIdToLocation`.

So we could support AH foreign assets to be forwarded to PC as well in the future using the same pallet.

pallets/assets-forwarder/src/lib.rs
```diff
@@ -344,6 +355,33 @@ pub mod pallet {
...
 		}
+
+		/// Remove the [`ForwardedAssets`] entry of asset `id` and release its deposit to the
+		/// depositor.
+		///
+		/// Only [`Config::ManagerOrigin`] can call this. The removal is local: the replica on the
+		/// destination is untouched, so a later re-forward fails there while the replica exists.
+		#[pallet::call_index(2)]
+		#[pallet::weight(<T as Config<I>>::WeightInfo::remove_forwarded_asset())]
+		pub fn remove_forwarded_asset(
```
gui1117, line 366: Note that this doesn't cover the failure of `sync_asset_status`.

If we remove the forwarded asset, then the next call to `forward_asset` will re-recreate the entry and the XCM will fail, and `sync_asset_status` can't be called because the entry exists and is up to date.

EDIT: covering failure is good but also not really mandatory, XCM is not supposed to fail, and root can always fix it. And all those additional code adds complexity.

georgepisaltu:
Currently blocked on https://github.com/paritytech/polkadot-sdk/pull/13059

pallets/assets-forwarder/src/lib.rs
georgepisaltu: Once the associated sdk PR is merged, we will move to a design which supports foreign assets as well.

ggwpez:
You will need 2606-2 patches for the bumped crates to get this deployed^ ?

georgepisaltu:
> You will need 2606-2 patches for the bumped crates to get this deployed^ ?

I will need that PR to land in sdk, get that released in some patch/minor sdk release, then upgrade individuality to use that sdk version, then update this PR. However, it is not a priority for this upcoming release, this PR is optional.

pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+	use frame_system::pallet_prelude::*;
+
+	#[pallet::pallet]
+	pub struct Pallet<T, I = ()>(_);
+
+	#[pallet::config]
+	pub trait Config<I: 'static = ()>: frame_system::Config + pallet_assets::Config<I> {
```
gui1117, line 131: isn't the refactor about using the fungible that is aggregating multiple pallet asset instances? why do we depend on pallet_asset of a specific instance I?

EDIT: probably not changed yet.

georgepisaltu: It's not changed yet, I am waiting for the sdk release.

pallets/assets-forwarder/src/benchmarking.rs
```diff
@@ -0,0 +1,131 @@
...
+impl BenchmarkHelper for () {
+	fn open_destination_channel() {}
+}
+
+/// Creates a live asset and returns its parameter form, with `caller` funded to pay the forward
+/// deposit and the delivery fees.
+fn setup_asset<T: Config<I>, I: 'static>(
+	caller: &T::AccountId,
+) -> Result<T::AssetIdParameter, BenchmarkError> {
+	<T as Config<I>>::BenchmarkHelper::open_destination_channel();
```
agustinustheo, line 43: AFAICT benchmark setup will open an empty outbound queue right?

To test the worst case, could the setup seed an existing outbound page?

An existing page will result in an additional storage read.

Here's an example for this setup (although it creates a small page doesn't create the worst case):
```rust
fn open_destination_channel() {
    ParachainSystem::open_outbound_hrmp_channel_for_benchmarks_or_tests(
        PEOPLE_ID.into(),
    );

    send_xcm::<xcm_config::XcmRouter>(
        xcm_config::PeopleLocation::get(),
        Xcm(vec![ClearOrigin]),
    )
    .expect("benchmark destination channel must accept XCM");
}
```

pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+
+			Self::deposit_event(Event::ForwardRemoved { asset_id });
+			Ok(())
+		}
+	}
+
+	impl<T: Config<I>, I: 'static> Pallet<T, I> {
+		/// Returns this pallet's interior location, computed from its actual instance index so it
+		/// cannot drift from the runtime's `construct_runtime!` ordering. The destination's origin
+		/// filter must match this index.
```
agustinustheo, line 390:
```suggestion
		/// Returns this pallet's interior location from its runtime instance index.
		///
		/// The destination's origin filter must match this index.
```

pallets/assets-forwarder/src/lib.rs
```diff
@@ -0,0 +1,457 @@
...
+//! this pallet's location, so its `pallet-assets` instance must set an origin filter (for example
+//! `EnsureXcm<Equals<location>>`) that matches `(1, [Parachain(<this chain>), PalletInstance(<this
+//! pallet>)])` as `ForceOrigin`.
+//!
+//! The replica's owner and team resolve to an account nobody controls on the destination, so its
+//! issuance there can only change through XCM transfers backed by this chain. The caller pays the
+//! XCM delivery fees and a deposit which backs the [`ForwardedAssets`] entry; only
+//! [`Config::ManagerOrigin`] can release the deposit, by removing the entry with
+//! [`Pallet::remove_forwarded_asset`], which is what makes junk forwards costly. The removal is
+//! local and does not produce any effects on remote chains.
```
agustinustheo, line 33:
```suggestion
//! XCM delivery fees. A deposit backs the [`ForwardedAssets`] entry and increases the cost of
//! unsolicited forwards. Only [`Config::ManagerOrigin`] can release the deposit by calling
//! [`Pallet::remove_forwarded_asset`]. The removal is local and does not affect remote chains.

```

agustinustheo (approved):
nothing blocking

Zebedeusz (approved):

PR #72, opened 2026-08-27
pallet-resources: support dynamic allowances

### LINKED ISSUE #70: pallet-resources: The value of AccountsApiAllowance and NotificationAllowance cannot be changed without a migration
It would be practical to be able to change the value of `AccountsApiAllowance` and `NotificationAllowance` with the technical track, but it cannot be changed without a migration.

* The easiest solution is probably to change from increase/decrease to set/reset and not allow users to accumulate different allowance on the same account. But not sure if the new constraint is ok, for example when topping up allowance can require some overlap on the same account.
* Otherwise: store the allowance alongside the account somewhere.

gui1117:
Closes https://github.com/paritytech/individuality-community/issues/70

### DIFF
--- a/pallets/resources/src/lib.rs
+++ b/pallets/resources/src/lib.rs
@@ -23,11 +23,6 @@
 //! period. A notification allowance authorizes a statement account to publish a notification
 //! without revealing the member's identity.
 //!
-//! Every statement allowance grant records the value it added next to the granted account. A
-//! revocation subtracts the recorded value, so the configured allowances (`AccountsApiAllowance`,
-//! `NotificationAllowance`, `PersonStatementLimit`, `LitePersonStatementLimit`) can change while
-//! grants are active.
-//!
 //! # Deprecated: username management
 //!
 //! Only the username-related state and extrinsics in this pallet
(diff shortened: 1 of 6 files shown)


andrzejSulkowski (approved):

pallets/resources/src/lib.rs
```diff
@@ -982,7 +999,7 @@ pub mod pallet {
 				// Revoke the old allowance and clear the reverse lookup.
 				decrease_allowance_by(
 					existing.account_id.clone().into(),
-					T::AccountsApiAllowance::get(),
+					existing.allowance.into(),
```
georgepisaltu, line 1002: If someone stacks allowances on a single account (which is a valid usecase today in TruAPI), this will clear all previous allowances, making stacking impossible right?

georgepisaltu (commented):
I think we can get away with just changing the number of slots instead of allowance values. In the future, when we will want to change allowance values thanks to improvements in hardware/SSS, the simplest thing to do would be to enact the allowance value change only on the period boundary.

pallets/resources/src/lib.rs
gui1117: maybe, let's not merge it then. Let's make it not configurable by governance.

PR #75, opened 2026-08-27
Add batch PGAS claims

georgepisaltu:
This PR adds batch claims for PGAS in `pallet-pgas`. There is only one proof, verified in multiple contexts using existing multi io primitives. This is ok since the claim is happening in a single account and each claim will have its own context. The benchmark only measures the worst case for now and always charges the worst case.

For AH, 5 claims should be enough to set an alias account.


pallets/pgas/src/lib.rs
```diff
@@ -170,6 +184,12 @@ pub mod pallet {
 		#[pallet::constant]
 		type MaxPgasClaimRecordCleanupPerCall: Get<u32>;
 
+		/// Maximum number of claim slots a single [`Call::batch_claim_pgas`] may carry.
+		/// Must not exceed [`MAX_PGAS_BATCH_CLAIMS`] or the largest per-collection claim
+		/// count; both asserted in `integrity_test`.
+		#[pallet::constant]
+		type MaxPgasClaimsPerBatch: Get<u32>;
```
gui1117, line 194: we could add in the doc that this value must be constant for the runtime. As it defines the transaction type and is benchmarked against.

Just so that it doesn't end up in pallet-parameters by mistake.

We could also enhance the code so that it supports run-time change.

pallets/pgas/src/extension.rs
```diff
@@ -232,6 +251,94 @@ impl<T: Config> AsPgas<T> {
...
+		)
+		.map_err(|_| InvalidTransaction::BadProof)?;
+		let day_be = Day::from(day);
+
+		let mut aliases = BatchAliases::new();
+		let mut validity =
+			ValidTransaction::with_tag_prefix("Pgas:Claim").priority(tx_priority::USER_DEFAULT);
+		for (ca, slot_index) in cas.iter().zip(slot_indices.iter()) {
+			let alias = ca.alias;
+			// Pool-hygiene pre-check; dispatch re-checks authoritatively.
```
gui1117: it is more than just a pool hygiene it prevents already claimed alias to spam the chain itself with unpaid failing transactions.

pallets/pgas/src/lib.rs
```diff
@@ -386,6 +431,44 @@ pub mod pallet {
...
+			alias: Alias,
+			day: Day,
+			collection: PgasCollection,
+			target: &T::AccountId,
+		) -> DispatchResult {
+			ensure!(!ClaimedGasAliases::<T>::contains_key(day, alias), Error::<T>::AlreadyClaimed);
+
+			let amount = T::PgasClaimAmount::get();
+			T::Fungibles::mint_into(T::PgasAssetId::get(), target, amount)
+				.map_err(|_| Error::<T>::PgasMintFailed)?;
```
gui1117, line 463: if it fails, then the transaction can be re-submitted on and on. But I think we can assume that it doesn't happen.

PgasClaimAmount should be higher than E.D. and PGAS is sufficient.

gui1117 (approved):

pallets/pgas/src/lib.rs
georgepisaltu: It should not fail with proper configuration. I think the other claim extrinsic suffers from the same issue, but I think it's ok for now.

pallets/pgas/src/extension.rs
georgepisaltu: Rephrased.

pallets/pgas/src/lib.rs
georgepisaltu: Added some doc explanations.

Zebedeusz (approved):

PR #74, opened 2026-08-27
improve jobs for first release

kianenigma:
See [RELEASE.md](https://github.com/paritytech/individuality-community/pull/74/changes#diff-2b1b69303b927a484e02c7fad9fc87d0d3ff0dc22ae1da0ecd0dc935d922a23c)

The plan after this would be to release with runtime spec versions `3_000_000` and version `0.3.0`.

* https://github.com/paritytech/individuality-community/pull/78 can integrate to the new flow once it is merged. 
* https://github.com/paritytech/individuality-community/pull/66 can be closed if @mordamax approves this.

PR #78, opened 2026-08-28
Gate releases against previewnet and paseo-next-v2

mordamax:
Runs the Polkadot proof-of-personhood chain tests against every release cut here, on a fork of each network these runtimes are actually deployed to.

Today a runtime is cut and handed to SRE with nothing having tried it against a real network first. This closes that: by the time a release leaves this repo it carries a verdict.

## What it does

Adds one `chain-tests` job to **Release** and **Nightly Release**. It calls the gate in [paritytech/polkadot-pop-e2e](https://github.com/paritytech/polkadot-pop-e2e), which forks previewnet and paseo-next-v2, upgrades each to these runtimes, and runs the suite — chain health, ring builder, identity backend, allowances, PoP-gated contract calls, alias claims.

It **reuses the wasm `build-runtimes` already produced** rather than rebuilding. A called workflow's jobs run inside this run, so the gate downloads the artifacts directly — no token for that part, no cross-repo API, no second srtool build.

Each network is diffed against its own deployed pin, so the run reports what would actually change:

```
asset-hub: runtime paritytech/individuality@v0.12.1-previewnet → <this build>
people:    runtime paritytech/individuality@v0.12.1-previewnet → <this build>
```

## It does not block publishing

The publish jobs don't depend on it. A gate failure reds the run and tells you what broke without holding up artifacts. Add `chain-tests` to publish's `needs` to make it blocking — worth doing once the signal has proven itself, not before.

## The one secret it needs

**`GH_PAT_PPNV1`** — `contents: read` on `paritytech/preview-net-v1`. Set.

That is the private engine that spawns the fork, and it is the *only* private repo involved. The runtimes come from the artifact built above rather than being fetched from `paritytech/individuality`, and every other pin — bulletin, web3-storage, release-automation, paseo-network/runtimes, kubo — is public and needs no token.

If the token is ever wrong, the gate fails fast and names the repo it could not read rather than dying on an unexplained 403 later.

## spec_version

A build straight from source carries the committed `spec_version`, which is what the chain already runs — so the first real run reported `2000039 -> 2000039 (already-installed)`, upgraded nothing and tested none of the PR's code while going green.

The build now stamps `+1` in the workspace only, landing in the band [RELEASE.md](RELEASE.md) reserves (`M_XXX_YYY`, `YYY` always `000` on main and belonging to nightlies). Nothing is committed back, so the `Spec version` check — which reads the committed source — is unaffected, and the value stays clear of every future release.

Reads whatever is committed, so it follows #74's move to `3_000_000` with no change here.

## Cost

Two forks per release, roughly 25 minutes each on `parity-large`, in parallel with publishing. Nightly and tagged releases only — nothing on PRs.

Gating a PR's runtime before merge is possible with the same machinery (the gate takes a built wasm, not just a published release) but means an srtool build per PR, so it's a separate conversation.


mordamax:
Rebased the spec_version approach onto #74's scheme: the build now stamps `+1` rather than `* 100`, landing inside the `YYY` band RELEASE.md reserves. It reads whatever is committed, so it follows the move to `3_000_000` with no change here.

Happy to close #66 in favour of #74 — this integration doesn't depend on it.

mordamax:
@paritytech/pop i made a release/nightly publish depend on the gate - so if chain-tests fails - the release doesn't get cut. Before it would publish anyway and just leave a red badge

but that means a failed release or nightly can now go unnoticed - nothing gets published and nobody is told

so i'd suggest we post a notification to your team room when it fails, so it's visible and someone picks it up (i can wire it up with our Earvin matrix bot - just tell me the room ID where you want to get notification and preferred format - otherwise i'll improvise))

p.s. auto issues i'd avoid - easy to ignore and they pile up

andrzejSulkowski:
> so i'd suggest we post a notification to your team room when it fails, so it's visible and someone picks it up (i can wire it up with our Earvin matrix bot - just tell me the room ID where you want to get notification and preferred format - otherwise i'll improvise))

I think this is a great idea! You are by now inside our elements group chat. For the format I can imagine something like "[WARNING] E2E Release Tests Failed (ref)\n${maybe title of failed test?}".  We could as well make it a [CRITICAL] as it blocks the release.

.github/actions/build-e2e-runtime-wasm/action.yml
```diff
@@ -49,17 +49,48 @@ runs:
...
+      run: |
+        set -euo pipefail
+        for lib in runtimes/next-asset-hub-paseo/src/lib.rs runtimes/next-people-paseo/src/lib.rs; do
+          # sed captures only the number: a bare [0-9_]+ grep also matches the
+          # underscore inside "spec_version" itself.
+          BASE=$(sed -nE 's/.*spec_version:[[:space:]]*([0-9_]+).*/\1/p' "$lib" | head -1 | tr -d '_')
+          [ -n "$BASE" ] || { echo "::error::no spec_version found in $lib"; exit 1; }
+          STAMP=$((BASE + 1))
+          GROUPED=$(sed -E 's/(.{1,3})(.{3})(.{3})$/\1_\2_\3/' <<< "$STAMP")
+          sed -i -E "s/spec_version:[[:space:]]*[0-9_]+/spec_version: ${GROUPED}/" "$lib"
```
andrzejSulkowski: removes comments `//` from selection
```suggestion
          BASE=$(sed -nE 's/^[[:space:]]*spec_version: ([0-9][0-9_]*),$/\1/p' "$lib" | head -1 | tr -d '_')
          [ -n "$BASE" ] || { echo "::error::no spec_version found in $lib"; exit 1; }
          STAMP=$((BASE + 1))
          GROUPED=$(sed -E 's/(.{1,3})(.{3})(.{3})$/\1_\2_\3/' <<< "$STAMP")
          sed -i -E "s/^([[:space:]]*spec_version: )[0-9][0-9_]*,$/\1${GROUPED},/" "$lib"
```

.github/workflows/nightly-release.yml
```diff
@@ -95,9 +95,30 @@ jobs:
...
+  # The gate forks the live network, upgrades it to these runtimes, and runs the
+  # suite. It reuses the artifacts from build-runtimes rather than rebuilding:
+  # a called workflow's jobs run in this run, so it can download them directly.
+  #
+  # Blocking: publish-nightly waits on this. A runtime that cannot be upgraded
+  # onto a fork of the networks it is headed for should not be published, and a
+  # gate failure that only reds the badge is easy to miss.
+  chain-tests:
+    needs: [build-runtimes]
+    uses: paritytech/polkadot-pop-e2e/.github/workflows/release-gate.yml@main
```
andrzejSulkowski, line 114: `polkadot-pop-e2e` is executing with write permission on our repo. Should we pin it to a specific commit to prevent unexpected changes? I think its generally a nice security win as well, even tho we own `polkadot-pop-e2e`.

.github/actions/build-e2e-runtime-wasm/action.yml
```diff
@@ -49,17 +49,48 @@ runs:
       with:
         shared-key: individuality-cache-e2e-runtime-wasm
 
+    # A runtime upgrade only runs when spec_version increases, and a build from
+    # source carries the value the chain already has. +1 lands in the YYY band
+    # RELEASE.md reserves for nightlies. Workspace only; nothing is committed.
+    - name: Stamp a strictly-greater spec_version
+      run: |
+        set -euo pipefail
```
andrzejSulkowski, line 62: we could reuse here ` LATEST=$(.github/scripts/latest_release_spec_version.sh)` which pulls in the actual latest spec version, not only the committed one.

andrzejSulkowski (approved):
Only the couple of comments. Rest looks good

.github/actions/build-e2e-runtime-wasm/action.yml
mordamax: good catch, taken. anchored both the read and the write to the exact line now, so a commented out spec_version cant match

.github/actions/build-e2e-runtime-wasm/action.yml
mordamax: done, it now stamps above max(committed, newest release) + 1

this actually bit us already - the branch had 2_999_000 while previewnet was on 3_000_000, so the gate refused it as a downgrade. with the release floor that cant happen

.github/workflows/nightly-release.yml
mordamax: added `permissions: contents: read` on the job instead, which i think gets you the actual win here - a called workflow inherits the callers token, and release.yml grants contents:write + pull-requests:write. now it gets read only

on pinning: it wouldnt do what you expect today. `uses: ...@sha` picks the workflow file at that sha, but the gate checks out polkadot-pop-e2e@main inside itself, because job_workflow_ref/job_workflow_sha come back empty in the github context (measured, theyre OIDC only). so youd pin the yaml and still run main s code. to really pin it you also pass `gate_ref: <sha>`, and then every update needs both bumped

happy to do that if you want it, just didnt think it was worth the maintenance unless you see it as a blocker

mordamax:
@andrzejSulkowski i've followed up with fixes and matrix room message, I'll need to invite my bot there too then )

.github/workflows/nightly-release.yml
andrzejSulkowski: I like your solution better.

andrzejSulkowski (approved):

tindzk (approved):

PR #79, opened 2026-08-31
coinage: benchmark generalization.

gui1117:
When `MinimumExponentForOutputUnloadFee` is below the unload fee the benchmark fails, which requires to set a specific config only so the benchmark succeeds. The benchmark now supports any config.


tindzk (approved):

agustinustheo (approved):

agustinustheo:
After looking at the merge diffs, this PR is no longer needed and is stale after https://github.com/paritytech/individuality-community/pull/99 is merged. Closing.

PR #80, opened 2026-08-31
Coinage: make parameters changeable at run time.

gui1117:
Just remove the `#[pallet::constant]` and move it to pallet view.


georgepisaltu (approved):

kianenigma (approved):

pallets/coinage/src/lib.rs
```diff
@@ -1172,22 +1174,25 @@ pub mod pallet {
...
 		type UnloadTokenAllowancePerTimePeriodForLitePeople: Get<NativeBalanceOf<Self>>;
 
 		/// Hard upper bound on the number of free unload tokens per time period.
 		///
 		/// The effective free token limit is:
 		/// `min(allowance / current_fee, MaxFreeUnloadTokensPerTimePeriod)`.
-		#[pallet::constant]
+		///
+		/// This parameter can be changed at any time.
 		type MaxFreeUnloadTokensPerTimePeriod: Get<u32>;
```
ggwpez, line 1196: Why not pub?

ggwpez (approved):

pallets/coinage/src/lib.rs
gui1117: it is an associated type, not a regular type.

gui1117:
not sure what label is appropriate:
* it doesn't require a migration
* it doesn't require host function
* it doesn't require a new config types
* but it does require users to adapt their usage of the runtime.

kianenigma:
good point, need to think of a better label for this @gui1117 .

PR #83, opened 2026-08-31
Re-benchmark post EC HF disablement 

### LINKED ISSUE #65: Regenerate Runtime Weights
follow up on #63.

This makes all proof based paths slower. Additionally we did a sdk bump to sdk [2606](https://github.com/paritytech/individuality-community/pull/10)

kianenigma:
closes https://github.com/paritytech/individuality-community/issues/65


kianenigma:
@gui1117 the job timed out after 24h

Let's retry this after https://github.com/paritytech/individuality-community/pull/91

gui1117:
I think we should benchmark with coinage excluded, and coinage should use either `steps=2` or `steps=4` and `repeat=10`.
`steps=50` is too much for coinage, I am doing another approach to completely remove the issue with steps and coinage at the moment so that in the future we don't have this problem anymore.

PR #84, opened 2026-08-31
Implement private NFT claims

tindzk:
Introduce a private claim path for attestation NFTs to break the link between the player who earned
an NFT claim credit and the account that mints it. Public claims prove a Merkle leaf built from the
claimant, so the mint names the player. A private claim proves ring VRF membership of one game's
registrants, showing that some registrant earned the credit but not which one.

A game opts in at scheduling time and is then private only. Each claimant registers a one-time ring
VRF key for one credit, and the offchain worker builds a ladder of rings once registration closes,
ring `t` holding every registrant that earned at least `t` credits. The ladder goes to the claims
chain over XCM, where a claimant spends one nullifier per tier they earned through `claim_private`,
which carries no signer and pays no fee. The claims run in one window, which opens
`PrivateClaimDelay` after the ladder arrives and closes `PrivateClaimWindow` later, and a member
who misses it mints nothing.

The ladder stops at the first tier whose ring is too small to hide its claimants, forfeiting the
credits above it. A game whose ring 1 falls short is abandoned and mints over the public path.

## Diagram
```mermaid
sequenceDiagram
    participant C as Claimant wallet
    participant P as People chain (NftCredits)
    participant O as Offchain worker (People)
    participant A as Asset Hub (NftClaims)
    participant OA as Offchain worker (Asset Hub)

    Note over P: game scheduled with claims: Private
    Note over P: game starts, player process runs
    P->>P: first credit award writes the game's private record, one tier counter per credit it can award
    P->>P: each award moves the claimant up one tier, which is the tier their key registers into
    Note over P: player process ends, credits final, key registration opens for PrivateKeyRegistrationSeconds
    C->>P: register_private_claim_key(game, one-time ring key), at the tier of their credit count
    Note over P: key registration closes
    O->>P: build_private_ring (zero keys, opens the ladder and decides its height)
    alt tier 1 below the anonymity floor
        O->>P: build_private_ring (zero keys, abandons the game)
    else tier 1 at or above the floor
        loop for each tier, from the top down
            loop PrivateKeysPerBuild of the tier's own keys per call
                O->>P: build_private_ring
            end
            O->>P: build_private_ring (zero keys, snapshots the tier's root)
        end
    end
    O->>P: send_private_ring
    P-->>A: XCM receive_private_rings, a ladder of roots or an abandonment
    O->>P: clean_up_private_game, after the delivery
    alt ladder
        A->>A: store the ladder, one row per tier, open the window after PrivateClaimDelay
        loop one call per tier the claimant earned, in an order the wallet picks
            C->>A: claim_private(game, tier, alias, proof, collection, mint_to)
            A->>A: authorize: verify the proof against the tier's root, the alias is unspent, the window is open
            A->>A: dispatch: spend the alias, mint into a fresh purse key
        end
        Note over A: window closes after PrivateClaimWindow
        OA->>A: close_private_ring, 32 aliases per call, then the ladder
    else abandonment
        A->>A: record the game as ended, which reopens the public claim path
    end
```


georgepisaltu:
After looking at the changes in this PR, I don't think we have a good way to offer private NFT claims today. If we had the blockspace to use Bandersnatch rings for these NFT credits, we could have sent over the roots of these rings to AH instead of the Merkle root approach we have today. I believe this is just too expensive for us to do on the People Chain today, especially as it is happening in close proximity with the game where we need the blockspace to progress on the game state. I think we should drop this effort for now until we can do it more efficiently. Also, there is a nasty UX gap where your NFT claims could either be private or take the regular minting path, so even using this blockspace wouldn't completely solve our problem.

shawntabrizi:
@georgepisaltu this pr is meant to address the reality that we dont have enough blockspace.

By default, we sent via merkle tries.

We create a special opt in queue for nfts to be put into a ring. We only create one ring (or however many we want).

And we only when there is available blockspace, so these rings would not be created during the high load of the game, but later when the activity goes down.

So if you want a private NFT, it is possible, but it is not "on-demand", its whenever we have availability.

This i think is a reasonable compromise for those who are privacy focused and willing to wait.

PR #85, opened 2026-09-01
Track the number of unloaded aliases per recycler ring

TorstenStueber:
Reading how many coins a live recycler has given out meant scanning every `RecyclerAliasStates` entry under the ring's prefix and decoding each value. There was no stored aggregate.

This stores the count per ring. No migration: every alias state is written through `mark_alias_unloaded`, so a ring with alias states but no count predates the count and stays uncounted, while a ring with no alias state has had no unload and counts from zero.

## Changes

- Add `RecyclersUnloadedCount: StorageMap<_, Twox64Concat, (InstanceId, Denomination, RingIndex), u32, OptionQuery>`. Absent means the ring is not counted.
- Add `RecyclerManager::unloaded_count(instance_id, value, index) -> Option<u32>`. `None` means the caller has to scan.
- `mark_alias_unloaded` increments the count; new `mark_unloaded_alias_locked(instance_id, value, index, alias, lock)` writes the failed-dispatch lock and decrements it. The transaction extension's `post_dispatch` calls the latter instead of writing `RecyclerAliasStates` itself.
- `clean_unchecked` drops the count with the ring, and logs a defensive error if a counted ring disagrees with its unloaded aliases.
- Rename `has_pending_dust` to `has_alias_states`.
- Tests: failed dispatch and retry, repeated unloads of one alias, cleanup, and the three counted/uncounted cases.

### DIFF
--- a/pallets/coinage/src/benchmarking.rs
+++ b/pallets/coinage/src/benchmarking.rs
@@ -1552,6 +1552,9 @@ mod benches {
 				AliasState::Unloaded,
 			);
 		}
+		// Keep the counter in step with the entries written above; `clean_unchecked` compares the
+		// two.
+		RecyclersUnloadedCount::<T>::insert((INSTANCE_ID, value, ring_index), m);
 
 		// Advance time past expiration
 		let status = T::MemberService::ring_status(&identifier, 0).expect("ring exists");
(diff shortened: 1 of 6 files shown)


TorstenStueber:
/cmd bench --runtime next-people-paseo pallet-weights --pallet indiv_pallet_coinage

gui1117 (approved):

kianenigma (approved):

kianenigma:
@TorstenStueber will you still benefit if we merge this, or will this be needed in a later release of TruApi/Host?

gui1117:
> @TorstenStueber will you still benefit if we merge this, or will this be needed in a later release of TruApi/Host?

you mean on next runtime upgrade?

PR #88, opened 2026-09-01
Let lite people register and claim people-airdrops draws

peetzweg:
Lets lite people participate in `PeopleAirdrops` draws: whitelists the people-airdrops context for lite account aliases and widens the pallet's person origin to accept lite aliases.

**Why:** The pallet is meant for every proven person, but on previewnet only full people can play. #48 registered the airdrops context in the *full-people* `AccountContexts` and its integration test exercised a full person, so the two lite-side gates were never hit: the lite pallet keeps its own context whitelist (`LitePeopleAccountContexts`, still auth + score only), and `EnsurePerson` was `EnsurePersonalAliasInContext`, which only matches the full-people `PersonalAlias` origin — even a bound lite alias dispatches to `BadOrigin`. Both gates confirmed live on previewnet with a garbage-proof probe: `set_alias_account` in the lite auth context fails `BadProof` (gate passed, proof checked), in the airdrops context it fails `Invalid::Call` (context gate).

**How:** Two runtime changes in `people.rs`: add `people_airdrops_context()` to `LitePeopleAccountContexts`, and set `EnsurePerson = EitherOf<EnsurePersonalAliasInContext, EnsureLiteAliasInContext>` — both yield `Success = Alias`, so the plain combinator suffices; full people are tried first. Aliases of the two tiers come from distinct member keys/collections, so they are disjoint registration entries; a person holding both identities registers as two aliases, inherent to the two-tier design. Other gates audited and already open: `AsLiteAliasWithAccount` re-checks only binding freshness (no context gate), `AsLiteAliasWithAccountRevised`'s context gate is covered by the same whitelist entry, and origin-restriction already grants `LiteAlias` a nonzero allowance.

Adds `lite_person_registers_and_claims_a_draw` and `lite_alias_account_in_another_context_cannot_register` mirroring the full-person tests; all 74 runtime integration tests pass, `runtime-benchmarks` check clean.

### DIFF
--- a/runtimes/next-people-paseo/src/integration_tests/people_airdrops_flow.rs
+++ b/runtimes/next-people-paseo/src/integration_tests/people_airdrops_flow.rs
@@ -83,6 +83,40 @@ fn bind_alias_account(secret: &VrfSecret, context: Context, account: &sr25519::P
 	exec_as_alias_with_proof(secret, context, set_alias.into());
 }
 
+/// Bind `account` to the lite person's alias in `context`, the lite-tier counterpart of
+/// [`bind_alias_account`].
+fn bind_lite_alias_account(secret: &VrfSecret, context: Context, account: &sr25519::Pair) {
+	let set_alias = indiv_pallet_people_lite::Call::<Runtime>::set_alias_account {
+		account: pair_to_account_id(account),
+		valid_at_block: frame_system::Pallet::<Runtime>::block_number(),
+	};
+	let uxt = build_as_lite_alias_with_proof_ext(secret, context, set_alias.into());
+	Executive::apply_extrinsic(uxt)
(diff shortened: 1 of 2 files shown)


peetzweg:
**Update: not needed.** The daily draws are intentionally limited to full (proven) people. The current behaviour on `main` is correct: lite people cannot bind an airdrop alias and the airdrop origin only accepts full-people aliases. Closing.

For the record, the live previewnet probe confirmed the two gates behave as designed: a lite person can bind an alias in the score context, and the same proof in the people-airdrops context is rejected with `Invalid::Call`.

peetzweg:
Closing: daily draws are meant for full (proven) people only, so the current origin and context whitelist are correct as they are.

PR #94, opened 2026-09-02
Personhood precompile: Refuse delegate calls and attached value

tindzk:
The personhood precompile served any incoming frame. It now refuses two scenarios before dispatch:

- a delegate call: the precompile would run against the delegator's address and storage
- attached value: the precompile address has no way to return

Both revert, so a mistaken caller keeps its forwarded gas and can catch the failure. No read-only guard is added, because `personhoodStatus` and `personhoodInfoByProof` are both views that a read-only frame has to serve.


0xRVE (approved):
LGTM

sphamjoli:
@tindzk im still reviewing

precompiles/personhood/src/tests.rs
```diff
@@ -298,3 +299,282 @@ fn proof_oversized_returns_none_and_refunds_gas() {
...
+	/// Register `account`'s address mapping with only the funds its deposit needs.
+	///
+	/// `map_account` endows `u64::MAX / 2`, and two such accounts plus a contract's deposits
+	/// overflow the mock's total issuance.
+	fn map_modestly(account: &AccountId32) {
+		Balances::make_free_balance_be(account, 1u64 << 40);
+		let _ = <Test as pallet_revive::Config>::AddressMapper::map(account);
+	}
+
+	alloy::sol! {
```
sphamjoli, line 371:
```suggestion
	// `ProofVerificationRequest` mirrors the struct in `sol/IPersonhood.sol`, the ABI source of
	// truth compiled by `alloy::sol!("sol/IPersonhood.sol")` in lib.rs. Keep the fields and their
	// order in step with that file: a mismatch mis-encodes `verifyProofInStaticFrame`, and this
	// test would pass while checking the wrong bytes.
	alloy::sol! {
```

precompiles/personhood/sol/IPersonhood.sol
```diff
@@ -27,6 +27,12 @@ pragma solidity ^0.8.30;
 ///   IPersonhood personhood = IPersonhood(0x000000000000000000000000000000000a010000);
 ///   IPersonhood.PersonhoodInfo memory info = personhood.personhoodStatus(someAddress, bytes32("dotns"));
 ///   if (info.status == 2) { /* full person */ }
+///
+/// Two conditions apply to every function and are not repeated below: native value attached to a
+/// call reverts, and a delegate call reverts. Both functions are views, so a read-only frame
+/// serves them.
+/// @custom:reverts "this precompile does not accept value"
+/// @custom:reverts "illegal to call this pre-compile via delegate call"
```
sphamjoli, line 35:
```suggestion
/// @custom:reverts "illegal to call this pre-compile via delegate call"
/// @dev The delegate-call revert reason comes from pallet-revive (`PrecompileDelegateDenied`).
///      Update it here if that upstream wording changes.
```

pallet-revive owns this string (`PrecompileDelegateDenied`). The value string above it is ours, so it stays correct.

precompiles/fixtures/contracts/PersonhoodCaller.sol
```diff
@@ -0,0 +1,90 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pragma solidity ^0.8.30;
+
+import "../../personhood/sol/IPersonhood.sol";
+
+/// @title PersonhoodCaller
+/// @notice Queries the personhood precompile four ways: a read and a proof verification inside a
+/// read-only frame that both succeed, a delegate call the precompile refuses, and a call that
```
sphamjoli:
```suggestion
/// read-only frame that both succeed, a delegate call the precompile refuses and a call that
```

House style drops the comma before "and" (`agents/style.md`).

sphamjoli (changes requested):
Requesting changes.

precompiles/personhood/src/lib.rs
sphamjoli, line 216:
```suggestion
					Err(_) => {
						log::error!(
							target: LOG_TARGET,
							"personhoodInfoByProof: failed to SCALE-decode proof bytes",
						);
						env.adjust_gas(charged, Weight::zero());
						return Ok(IPersonhood::personhoodInfoByProofCall::abi_encode_returns(
							&false,
						));
					},
```

sphamjoli (changes requested):

precompiles/personhood/src/lib.rs
tindzk: This would break `proof_precompile_decode_failure_charges_full_weight`.

Charging on decode failure protects against DoS. The other paths refund because they reject before any decode work.

sphamjoli (approved):
LGTM

Zebedeusz (approved):

PR #99, opened 2026-09-04
Fix coinage benchmark `as_unload_token_from_output_tx_ext`

andrzejSulkowski:
benchmark `as_unload_token_from_output_tx_ext` wanted to unload a coin which's value could not cover the unload fees throwing `UnloadedValueBelowFee`

### DIFF
--- a/pallets/coinage/src/benchmarking.rs
+++ b/pallets/coinage/src/benchmarking.rs
@@ -3965,9 +3965,8 @@ mod benches {
 	fn as_unload_token_from_output_tx_ext() -> Result<(), BenchmarkError> {
 		common_setup::<T>();
 
-		let value = denomination_covering_unload_fee::<T>(
-			T::MinimumExponentForOutputUnloadFee::get()
-		);
+		let value =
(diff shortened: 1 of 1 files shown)


pallets/coinage/src/benchmarking.rs
```diff
@@ -3957,8 +3965,8 @@ mod benches {
 	fn as_unload_token_from_output_tx_ext() -> Result<(), BenchmarkError> {
 		common_setup::<T>();
 
-		// Setup recycler with a denomination large enough for the penalty fee.
-		let value = T::MinimumExponentForOutputUnloadFee::get();
+		let value =
+			denomination_covering_unload_fee::<T>(T::MinimumExponentForOutputUnloadFee::get());
```
TorstenStueber: We can redefine `MinimumExponentForOutputUnloadFee` as `ConstI8<0>` again instead of using an (arbitrary) hard-coded `ConstI8<4>`.

TorstenStueber (approved):
Looks good! Thanks! Unloading became more expensive with the new change to increment the unload counter.

ggwpez (approved):

PR #102, opened 2026-09-06
ci: Fix test-runtime-upgrade job

csmnprschv:


andrzejSulkowski:
I was wondering if we should bump the spec version pro actively. When doing a release:
1. release with current spec version
2. bump spec version

Effect is that all changes go into spec version which is not released yet.
This seems to make more sense as then we dont need to manually bump when testing against the live network, which has the same spec version.

What do you think @csmnprschv?

tindzk (approved):

agustinustheo (approved):

PR #103, opened 2026-09-08
Retain credit awards on a wall-clock TTL

tindzk:
Awards are kept for the last 256 award blocks (`MaxRetainedAwardBlocks`). The persistence window therefore gets shorter the busier the chain: It is 8.5 minutes if every block awarded a credit given a block time of 2s. Asset Hub gives a claimant 90 days to mint against a root, so a credit could outlive the award.

Awards are now kept for a fixed length of time, matching that same 90-day claim deadline. Both chains now group what they expire into buckets of a day, keyed by the timestamp the tree commits to. Each sweep walks one bucket per call from
the oldest bucket it has not swept yet. This replaces the ordered map of per-second deadlines the roots and the claims chain's trees were filed in.

The trade-off is that the space used now grows with the number of people. About 13 KB per player at one game a week, and roughly 7 MB in the worst case of a player joining every game while games run back to back.

Fixes https://intelligence.security.parity.io/finding/individuality-1249-1786709755/6168


Zebedeusz (approved):

tindzk:
/cmd bench --pallet indiv_pallet_nft_credits indiv_pallet_game indiv_pallet_nft_claims

pallets/nft-credits/src/migration.rs
```diff
@@ -72,6 +110,20 @@ pub mod v1 {
...
 			}
+
+			let retained = alloc::vec::Vec::<BlockNumberFor<T>>::decode(&mut &state[..])
+				.map_err(|_| "retained tree blocks must decode")?;
+			for block in &retained {
+				ensure!(
+					NftClaimCreditAwardExpiries::<T>::iter().any(|(_, filed, _)| filed == *block),
+					"a retained tree block has no expiry entry"
+				);
+			}
```
agustinustheo, line 121: Read the expiry map once by using a set.

```suggestion
			let filed = NftClaimCreditAwardExpiries::<T>::iter_keys()
				.map(|(_, block)| block)
				.collect::<alloc::collections::BTreeSet<_>>();
			for block in &retained {
				ensure!(filed.contains(block), "a retained tree block has no expiry entry");
			}
```

pallets/nft-credits/src/tests.rs
```diff
@@ -3022,6 +3028,254 @@ mod root_removal {
...
+		new_test_ext().execute_with(|| {
+			record_tree_blocks(&[10]);
+
+			assert!(NftClaimCreditAwardExpiries::<Test>::contains_key(
+				ExpiryTimestamp::from(timestamp_of(10)),
+				10
+			));
+			assert_eq!(oldest_filed(), Some(timestamp_of(10)));
+		});
+	}
```
agustinustheo, line 3123: Use the helper that submits game reports and builds the tree. Then check the resulting expiry entry. The test must fail if tree building does not register the expiry.

```suggestion
	#[test]
	fn a_tree_block_is_filed_under_the_timestamp_of_its_root() {
		new_test_ext().execute_with(|| {
			let (block, _) = award_credits_in_one_block();
			let root =
				NftClaimCreditRoots::<Test>::get(block).expect("the block awarded credits");

			assert!(NftClaimCreditAwardExpiries::<Test>::contains_key(
				ExpiryTimestamp::from(root.timestamp),
				block
			));
			assert_eq!(oldest_filed(), Some(root.timestamp));
		});
	}
```

pallets/nft-credits/src/lib.rs
```diff
@@ -1800,6 +1832,38 @@ impl<T: Config> Pallet<T> {
 		Some(<T as Config>::WeightInfo::sweep_expired_roots(count)).into()
 	}
 
+	/// Removes the awards of up to [`Config::MaxAwardBlocksPerSweep`] tree blocks whose TTL has
+	/// run out, as [`Pallet::sweep_expired_awards`] does once its origin is checked.
+	///
+	/// The roots of those blocks stay: they run on the longer [`Pallet::root_ttl`], and a claimant
+	/// holding their own proof can still mint against one for as long as it does.
```
agustinustheo, line 1839: A bit longer explanation but I think it reads better to me:

```suggestion
	/// This sweep leaves the roots in place. Root expiry uses the tree's original timestamp plus
	/// [`Pallet::root_ttl`]. A claimant holding a proof can still mint while the claims chain
	/// retains the corresponding tree.
```

pallets/nft-credits/src/benchmarking.rs
```diff
@@ -42,15 +42,14 @@ mod benches {
...
-	// The ring is set up full, so the run includes dropping the oldest retained tree, which is the
-	// worst case and the one every block pays for once the chain has been running.
+	// Filing the block for expiry is a fixed pair of writes whatever `n` is, the awards being
+	// removed by a sweep of their own rather than by this path.
 	#[benchmark]
 	fn build_credit_tree(n: Linear<1, AWARDS_PER_TREE>) -> Result<(), BenchmarkError> {
-		let retained = T::MaxRetainedCreditTrees::get();
-		frame_system::Pallet::<T>::set_block_number((retained + 10).into());
+		frame_system::Pallet::<T>::set_block_number(10u32.into());
 		let block = frame_system::Pallet::<T>::block_number();
```
agustinustheo, line 53: I think this is not testing the worst case. In the worst case, the queue could be almost full with 255 entries and writing the 256th entry.

Also measure the full-queue rejection path to establish which path costs more.
```suggestion
		let capacity = T::MaxQueuedCreditTrees::get();
		queue_credit_trees::<T>(capacity.saturating_sub(1));
		frame_system::Pallet::<T>::set_block_number(capacity.saturating_add(10).into());
		let block = frame_system::Pallet::<T>::block_number();
```

pallets/nft-credits/src/benchmarking.rs
agustinustheo, line 83:
```suggestion
		assert_eq!(credit_root.leaf_count, n);
		assert_eq!(CreditTreeDeliveryQueue::<T>::decode_len(), Some(capacity as usize));
```

agustinustheo (approved):
please address the comments, otherwise ok

pallets/nft-credits/src/migration.rs
tindzk: Addressed.

pallets/nft-credits/src/tests.rs
tindzk: Addressed.

pallets/nft-credits/src/lib.rs
tindzk: Addressed.

pallets/nft-credits/src/benchmarking.rs
tindzk: Addressed.

pallets/nft-credits/src/benchmarking.rs
tindzk: Addressed.

PR #104, opened 2026-09-10
Additions to claude instructions based on polkadot-sdk developments

Zebedeusz:
Proposes additions to Claude instructions:
- in extrinsics.md mentions `#[frame_support::stored]` construct,
- adds precompiles.md.


agents/precompiles.md
```diff
@@ -0,0 +1,13 @@
...
+- Order the guards at the top of `call`, before any charge or state read: `ensure_not_delegate`, then `ensure_no_value`, then the read-only check.
+- `ensure_not_delegate` is unconditional. Under a delegate call the precompile runs with the delegator's address and storage, and `env.caller()` names the original caller, so identity and any address-derived id are attacker-controlled. Refuse even when no current selector derives identity from them, so a later one cannot introduce the hole silently.
+- `ensure_no_value` is unconditional for a non-payable precompile. The address has no owner, no code and no withdrawal path, so attached value is unrecoverable.
+- Both guards revert rather than trap, so the caller keeps its forwarded gas and can catch the failure.
+- Gate the read-only check on the selector, not on the whole precompile: `if env.is_read_only() && is_mutating(input)`, as in `precompiles/scarcity/src/collection.rs:51` and `precompiles/nft-claims/src/lib.rs:134`. A view selector must stay callable from a read-only frame. A blanket `env.is_read_only()` denial is correct only when every selector mutates, as in `precompiles/scarcity/src/factory.rs:45`.
```
tindzk: The line numbers may get outdated quickly. We could drop these parts:

```
, as in `precompiles/scarcity/src/collection.rs:51` and `precompiles/nft-claims/src/lib.rs:134

, as in `precompiles/scarcity/src/factory.rs:45`
```

agents/precompiles.md
```diff
@@ -0,0 +1,13 @@
...
+- Resolve the caller with `caller_account`, not by hand. It reverts with `ERR_INVALID_CALLER` when the EVM address has no mapped account id.
```
sphamjoli:
```suggestion
- Resolve the caller with `caller_account`, not by hand. It reverts with `ERR_INVALID_CALLER` when the caller is a `Root` origin, which has no account id.
```

The trigger is wrong. `caller_account` resolves the caller's account id, and that lookup fails only for a `Root` origin, which is what `ERR_INVALID_CALLER` reports. An unmapped address is a separate error the helper never reaches, and an EVM address maps through the fallback anyway. So a negative test written against an unmapped address would never revert. The same wording sits in the helper's own doc comment in `indiv-precompile-support`, so fix it there too.

agents/precompiles.md
```diff
@@ -0,0 +1,13 @@
+# Precompiles
+
+- Use `indiv-precompile-support` for the shared guards. Do not re-implement them per crate.
+- Order the guards at the top of `call`, before any charge or state read: `ensure_not_delegate`, then `ensure_no_value`, then the read-only check.
+- `ensure_not_delegate` is unconditional. Under a delegate call the precompile runs with the delegator's address and storage, and `env.caller()` names the original caller, so identity and any address-derived id are attacker-controlled. Refuse even when no current selector derives identity from them, so a later one cannot introduce the hole silently.
```
sphamjoli:
```suggestion
- `ensure_not_delegate` is unconditional. Under a delegate call the precompile runs with the delegator's address and storage, and `env.caller()` is the delegator's own caller (msg.sender is preserved), so identity and any address-derived id are attacker-controlled. Refuse even when no current selector derives identity from them, so a later one cannot reintroduce the risk.
```

Two wording fixes. Under a delegate call `env.caller()` is the delegator's own caller, one frame up, not the transaction origin, so **names the original caller** misreads it. And **introduce the hole silently** editorialises, which `style.md` asks us to avoid; **reintroduce the risk** states it plainly. The security point itself is right.

agents/precompiles.md
Zebedeusz: Done

tindzk (approved):

sphamjoli (approved):

andrzejSulkowski (approved):

PR #95, opened 2026-09-03
Coinage: improve benchmarks to handle many steps.

gui1117:
Typical CI use `steps=50` which makes the current benchmark of coinage taking too much time. Especially because the cache don't contain the proofs.

This PR change coinage so that instead of relying on ranges, we just rely on a pre-defined set of benchmark.

PR #107, opened 2026-09-13
ci: gate stale coinage benchmark proof caches

### LINKED ISSUE #106: CI: detect stale Coinage benchmark proof caches
Coinage benchmarks use cached ring-VRF proofs to avoid expensive proof generation during weight measurements. Changes to benchmark setup or proof inputs can invalidate that cache, causing subsequent benchmark runs to fall back to slow proof generation without an early CI failure.

Add a CI check that detects missing or stale proof-cache coverage when relevant Coinage benchmark inputs change. Validate coverage against the actual benchmark inputs rather than requiring a textual change to `proof_cache.rs`, since some benchmark changes legitimately leave the cached proofs unchanged.

## Changes

- Check that the committed proof cache covers the normal Coinage benchmark suite at the configured sampling settings.
- Fail with actionable regeneration instructions when required proofs are missing.
- Exclude diagnostic extra benchmarks from the normal cache-coverage requirement.
- Allow changes that preserve valid cache coverage without requiring an unnecessary cache-file modification.

Follow-up to the deferred review feedback on PR #95 by @tindzk:
https://github.com/paritytech/individuality-community/pull/95#discussion_r3923713337

agustinustheo:
Closes #106

Coinage benchmark proof-cache misses previously fell back to expensive ring-VRF proof generation. That could turn the normal runtime benchmark job into an hours-long run without a clear failure.

This adds a strict proof-cache mode for the real next-people-paseo benchmark runtime, regenerates the required cache coverage, and makes CI surface an actionable annotation when coverage drifts.

This branch is based on open PR #95

## Changes

- Add pallet and next-people-paseo runtime strict proof-cache features; regeneration takes precedence when both are enabled.
- Make strict cache misses fail immediately with the exponent, member count, cache key and regeneration instructions.
- Enable strict mode only for next-people-paseo in the existing `benchmark-runtime` job, tee benchmark output and emit a GitHub error annotation on a cache miss.
- Add a strict-mode pallet test and a dedicated lint-job invocation because all-features builds enable regeneration.
- Document CI coverage behaviour and regenerate the R2e10 proof cache for the normal benchmark suite.


pallets/coinage/src/benchmarking/scripts/regen_proof_cache.py
```diff
@@ -188,9 +188,9 @@ def main() -> None:
         help="Print the entry count but do not modify proof_cache.rs",
     )
     p.add_argument(
-        "--profile", default="production",
+        "--profile", default="release",
```
tindzk, line 195: If the reason for this change was build time, then `dev` should be even faster.

pallets/coinage/src/benchmarking.rs
```diff
@@ -290,12 +290,18 @@ fn setup_built_unload_recycler<T: Config>(
...
-#[cfg(feature = "benchmark-proof-cache-regenerate")]
+#[cfg(any(
+	feature = "benchmark-proof-cache-regenerate",
+	feature = "benchmark-proof-cache-strict"
+))]
 use alloc::string::String;
-#[cfg(feature = "benchmark-proof-cache-regenerate")]
+#[cfg(any(
+	feature = "benchmark-proof-cache-regenerate",
+	feature = "benchmark-proof-cache-strict"
```
tindzk: Perhaps change `strict` to `check`?

tindzk (approved):

pallets/coinage/src/benchmarking/scripts/regen_proof_cache.py
agustinustheo: No it wasn't for build time.

pallets/coinage/src/benchmarking.rs
agustinustheo: https://github.com/paritytech/individuality-community/pull/107/commits/75297201a565836118449c40dc84623724df0388

PR #125, opened 2026-09-15
XCM barrier requires explicit unpaid execution from Asset Hub

### LINKED ISSUE #115: Backports: People's XCM barrier
Part of #87

**Change:** People's XCM barrier has a custom rule that accepts any Asset Hub message without inspecting its instructions. Replace it with `AllowExplicitUnpaidExecutionFrom` over the trusted origins, which requires `UnpaidExecution` and re-checks the origin after any instruction that alters it. Messages failing those checks lose the unpaid path. 

-> Existing deny rules stay

**Runtime:** `next-people-paseo` · **Ref:** [xcm_config.rs#L217](https://github.com/paritytech/individuality-community/blob/68cab75a9e2f6cf3d7f79da4a05593e033e42d91/runtimes/next-people-paseo/src/xcm_config.rs#L217)

- [x] ✅ Backport
- [ ] ❌ Skip

**Reason:**

In my quest to find out if this is worth doing or not, I already implemented it. Solely for this reason + the code changes are minimal.

andrzejSulkowski:
Closes #115

People's XCM barrier had a `AllowAssetHubExecution` rule that admitted any message from Asset Hub on an origin equality check alone. Now it follows the same requirements as on mainnet


runtimes/next-people-paseo/src/xcm_config.rs
```diff
@@ -239,18 +224,18 @@ pub type Barrier = TrailingSetTopicAsId<
...
-					// Parent and its pluralities (i.e. governance bodies) get free execution.
-					AllowExplicitUnpaidExecutionFrom<ParentOrParentsPlurality>,
-					// The chain the credit trees are delivered to. It sends back the roots it is
-					// finished with. Unlike `AllowAssetHubExecution` above, this accepts only a
-					// message that asks for free execution.
-					AllowExplicitUnpaidExecutionFrom<Equals<NextAhLocation>>,
+					// Parent and its pluralities get free execution.
+					AllowExplicitUnpaidExecutionFrom<(
+						ParentOrParentsPlurality,
+						Equals<AssetHubLocation>,
```
Zebedeusz: Why do we need these 3 locations for one AH?

andrzejSulkowski: thanks for the great question. 
After you asked I realized I didnt fully understand this part of the code either and its indeed redundant to include `Equals<AssetHubLocation>` & `Equals<NextAhLocation>` because `AssetHubLocation == NextAhLocation`. We have inside our network only one AssetHub, so I deleted one of them.

By the 3th location - I assume - you meant `AssetHubPlurality`? This is a different kind of match statement. I am not 100% sure what Plurality is, but it seems like its not a match against AH root, but more like entities inside AH who dispatched XCMs against People Chain.

andrzejSulkowski: actually... Idk why we have `AssetHubLocation` and `NextAhLocation`... both are the same.
I'll create a quick follow up PR to unify them into one `AssetHubLocation`.

Edit: https://github.com/paritytech/individuality-community/pull/126

Zebedeusz (approved):

runtimes/next-people-paseo/src/xcm_config.rs
Zebedeusz: Okay, with that change it looks good to me.

Zebedeusz: Although a response to this comment should resolve our doubts: https://github.com/paritytech/individuality-community/pull/126/changes#r4037122523

georgepisaltu: I think there were 2 AH runtimes, just like there were 2 PC runtimes, the "classic" one mirroring the Polkadot variant and the "next" one with new features. This is indeed confusing. About the plurality, that's essentially different bodies that can form a plurality like OpenGov or treasury (but on Asset Hub). Even if we don't technically have a use case for these, this is upstream stuff from Polkadot so we should do it too.

georgepisaltu (approved):

PR #126, opened 2026-09-16
Next People Paseo: Unify AH and NextAH locations into one

andrzejSulkowski:
we have two locations `NextAhLocation` and `AssetHubLocation` inside `xcm_config.rs` both pointing to the same location.

Inside https://github.com/paritytech/individuality-community/pull/125#discussion_r4024094220, I decided to unify in a follow up PR


runtimes/next-people-paseo/src/xcm_config.rs
```diff
@@ -41,7 +41,7 @@ use parachains_common::{
 	},
 	TREASURY_PALLET_ID,
 };
-use paseo_runtime_constants::system_parachain::{ASSET_HUB_ID, COLLECTIVES_ID, NEXT_ASSET_HUB_ID};
```
Zebedeusz, line 44: @csmnprschv what are your thoughts on that change?
Is ASSET_HUB_ID pointing to the externally managed paseo and NEXT_ASSET_HUB_ID to the one internally managed? 
Since this runtime is "next", should we rather use the NEXT_ASSET_HUB_ID here?

PR #136, opened 2026-09-17
pallet-game: refactor sign-up checks

tindzk:
Closes https://github.com/paritytech/individuality/issues/230.

`GameAsInvited::validate` re-implemented every precondition that `sign_up_inner` checks on dispatch, so the two could drift. This extracts each check into a helper that both the extension and the dispatch call. It also enforces the invariant that a sign-up with an invite never fails after validation.


agustinustheo (approved):
approved here: https://github.com/paritytech/individuality/pull/1229

PR #140, opened 2026-09-18
Scarcity: Replay protection for failed dispatches

tindzk:
Closes https://github.com/paritytech/individuality/issues/1272.

`AsScarcity::post_dispatch_details` restored the NFT unchanged on a failed dispatch. The same signed transaction stayed valid for the rest of its mortality window, so anyone who had seen it could resubmit it once the backoff lock expired. Restore the NFT at the next state nonce instead, which retires the transaction that failed. A nonce already at its maximum stays put, where dispatch fails anyway.

The window is reachable in practice: the default mortal era with period 64 (64 × 6s = 384s) lasts far longer than the 60s lock that Paseo configures.

The docs argued the replay could not happen because callers must sign with an era shorter than `Config::LockPeriod`. Nothing enforces that, so state the replay rules as they now hold.


pallets/scarcity/src/extension.rs
```diff
@@ -29,31 +29,22 @@
...
-//!   third-party replay of the old transaction.
-//!
-//! A holder can also cancel an outstanding authorization at any time by moving the NFT, which
-//! increments its state nonce.
+//! Purse authorization is not account-nonce-based: a transaction stays valid while its purse
+//! holds the named instance at the named state nonce, and anyone who has seen it can submit it.
+//! Every dispatch moves that nonce on, a success by moving the NFT and a failure by restoring it
+//! at the next nonce, which retires every authorization signed for the old state. One that never
+//! executes stays valid until the NFT moves, so callers should sign mortal transactions to bound
+//! stale intent.
```
Zebedeusz, line 47: Can you simplify this comment please?

Zebedeusz (approved):
Otherwise looks fine

pallets/scarcity/src/extension.rs
tindzk: Improved the comment.

pallets/scarcity/src/tests.rs
```diff
@@ -1672,6 +1674,42 @@ fn failed_dispatch_restores_and_locks() {
...
+		assert_noop!(Scarcity::transfer(origin, 4), Error::<Test>::AddressOccupied);
+		post_dispatch(pre, Err(Error::<Test>::AddressOccupied.into()));
+
+		assert_invalidity(
+			validate_transfer_as(OWNER, 6, signed.clone())
+				.err()
+				.expect("the purse is locked"),
+			CustomInvalidity::NftTemporarilyLocked,
+		);
+		MockNow::set(60);
```
agustinustheo: Nit: the current test already checks stale-nonce rejection, but I think this would model replaying the failed transaction. Also, checking for a new auth succeeds would confirm that the old auth is the proper rejection reason.

```suggestion
		MockNow::set(60);

		// Clear the original destination so occupancy cannot prevent the replay.
		let (_, val, origin) = validate_burn(4).unwrap();
		let pre = prepare_burn(val, &origin);
		assert_ok!(Scarcity::burn(origin));
		post_dispatch(pre, Ok(()));

		assert!(!NftsByOwner::<Test>::contains_key(4));
		assert_state_mismatch(
			validate_transfer_as(OWNER, 4, signed).err().expect("the state nonce moved on"),
		);

		// The same call validates with an authorization for the current state.
		assert!(validate_transfer(OWNER, 4).is_ok());
```

agustinustheo (approved):

PR #143, opened 2026-09-18
nft-claims, scarcity: address preview and metadata batch audit findings

tindzk:
Closes https://github.com/paritytech/individuality/issues/1328.

Ported from https://github.com/paritytech/individuality/pull/1335.


shawntabrizi:
We will need to merge #145 first and then rebase

